PR #11, opened 2025-10-01
Fix missing panel contents after restore

### LINKED ISSUE #1: Move all offcanvas block into restored panels
Due to cynkra/dockViewR#51, only the visible panel(s) are restores successfully.

nbenn:
Resolves #1.

### DIFF
--- a/R/ui.R
+++ b/R/ui.R
@@ -174,54 +174,67 @@ show_hide_block_dep <- function() {
   )
 }
 
-show_block_panel <- function(id, add_panel = TRUE, session = get_session()) {
+show_block_panel <- function(id, add_panel = TRUE, immediate = FALSE,
+                             session = get_session()) {
 
-  stopifnot(is_string(id), is_bool(add_panel))
+  stopifnot(is_string(id), is_bool(add_panel), is_bool(immediate))
 
   ns <- session$ns
 
-  if (add_panel && !block_panel_id(id) %in% list_block_panels(session)) {
+  if (add_panel) {
     add_block_panel(id, session)
-  } else {
-    log_debug("selecting panel {block_panel_id(id)}")
-    dockViewR::select_panel(dock_id(), block_panel_id(id), session)
   }
 
   bid <- ns(id)
   pid <- block_panel_id(id, dock_id(ns))
 
-  log_debug("showing block {bid} in panel {pid}")
+  callback <- function() {
 
-  session$sendCustomMessage(
-    "show-block",
-    list(
-      block_id = paste0("#", bid),
-      panel_id = paste0("#", pid)
+    log_debug("showing block {bid} in panel {pid}")
+
+    session$sendCustomMessage(
+      "show-block",
+      list(
+        block_id = paste0("#", bid),
+        panel_id = paste0("#", pid)
+      )
     )
-  )
+  }
 
-  invisible(NULL)
+  if (immediate) {
+    callback()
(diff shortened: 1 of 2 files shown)


R/ui.R
```diff
@@ -180,8 +180,11 @@ show_block_panel <- function(id, add_panel = TRUE, session = get_session()) {
 
   ns <- session$ns
 
-  if (add_panel) {
+  if (add_panel && !block_panel_id(id) %in% list_block_panels(session)) {
```
nbenn, line 183: @DivadNojnarg this is an example of an issue I get with the `input`-based functions like `get_panels_ids ()`: I wanted to create a "smart" `show_block_panel()` function that auto-adds a panel if it does not already exists. But with the current form of `get_panels_ids ()` I can never be sure of the result I get. It might be that the panel actually is there, but just not visible yet in `input`. This lead me to https://github.com/cynkra/dockViewR/issues/49#issuecomment-3357133722.

nbenn: Given that what I suggest in https://github.com/cynkra/dockViewR/issues/49#issuecomment-3357133722 is not possible, we'll have to come up with something else.

nbenn:
@DivadNojnarg I tried implementing the pattern (1) restore, (2) iterate over panels & select each, to make sure they are available in the DOM and (3) insert some UI into the panel. So far it's a bit of a struggle due to async communication (and perhaps the way dockViewR goes about certain things; still trying to understand this better). Consider the following example.

First attempt:

```r
dockViewR::restore_dock("dock", layout, session)
for (id in names(layout[["panels"]])) {
  dockViewR::select_panel("dock", id, session)
}
```

This does not work b/c right after the "restore", the input validation relying on `get_panels_ids(.)` fails.

If I try to move the "selection" step to the end of a "flush-cycle", this also does not work. I tried something like

```r
dockViewR::restore_dock("dock", layout, session)
session$onFlushed(
  function() {
    for (id in names(layout[["panels"]])) {
      dockViewR::select_panel("dock", id, session)
    }
  },
  once = TRUE
)
```

Now I'm accessing `input$dock_state` outside of a reactive context. Next I try to `isolate()`

```r
dockViewR::restore_dock("dock", layout, session)
session$onFlushed(
  function() {
    for (id in names(layout[["panels"]])) {
      isolate(
        dockViewR::select_panel("dock", id, session)
      )
    }
  },
  once = TRUE
)
```

Now, even at the end on the next flush-cycle, I am still not seeing the new IDs. This surprised me a bit. I still don't have a solid enough understanding the mechanics here.

I can ofc wrap the "selection" in an `observeEvent()`, like so:

```r
dockViewR::restore_dock("dock", layout, session)
observeEvent(
  req(
    all(names(layout[["panels"]]) %in% dockViewR::get_panels_ids("dock"))
  ),
  {
    for (id in names(layout[["panels"]])) {
      dockViewR::select_panel("dock", id, session)
    }
  },
  once = TRUE
)
```

This works, but now I have another problem: I also have step (3), which must happen after (2), but constructing a solid trigger for that observer is getting increasingly more involved.

An important thing to note here: my whole construction might be unnecessary, especially the first part: If we didn't have the "input validation" step in `dockViewR::select_panel()`, at least some of the problems would go away, no? All I want is to execute a sequence of `session$sendCustomMessage()` calls. I'm assuming on the client side, they will be run in the order they were issued server-side (is this assumption correct?).

Issue #8, opened 2025-10-01
More verbose JS utils for moving DOM elements


nbenn:
What we currently have for this in

https://github.com/BristolMyersSquibb/blockr.dock/blob/00c4a3c5dc24cffa440bc93730d932018ed9906f/inst/assets/js/show-hide-block.js#L1-L12

silently fails if required DOM elements are not present/found. Depending on some log level, i.e. if `get_log_level() >= debug_log_level` it would be helpful if there were at least warnings in JS, maybe even server-side.

DivadNojnarg:
I believe we can close this one as done in #13. Now we have JS warnings.

PR #13, opened 2025-10-02
Dock extensions

### LINKED ISSUE #6: Extension mechanism
Create an extension mechanism to allow for extensions in panels.

nbenn:
Closes #6, #8.

### DIFF
--- a/R/ext-class.R
+++ b/R/ext-class.R
@@ -16,7 +16,7 @@
 #' @export
 new_dock_extension <- function(server, ui, name, class, ctor = sys.parent(),
                                pkg = NULL, ...) {
-	validate_extension(
+  validate_extension(
     structure(
       list(server = server, ui = ui, ...),
       name = name,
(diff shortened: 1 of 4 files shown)


nbenn:
Unsolved issues:
- How does an extension react to a board update? Some core hook in the update procedure?
- How can extensions be restored from JSON?

nbenn:
@DivadNojnarg I added a super basic "board editor" using the core UI/logic to experiment with. We can edit the board using this. Might still be a bit buggy, but we're getting there!

nbenn:
@DivadNojnarg as discussed, this is currently a bit WIP. What we need to wrap this up I believe is

- betters tools (JS) for moving around block/ext UI (maybe targeting a div around each block/ext that is there specifically for moving stuff around),
- a bit of UI cleanup for the `edit_board_extension` (mainly alignment issues),
- make sure that some basic things like app startup (empty and with some blocks and/or exts) works, adding/removing some blocks/links is ok and ofc restoring a saved board works

Then I think we can merge this. There are some things from here we should look into medium term:

- simplifying the `edit_board_extension` logic and maybe coming up with a bit nicer UI
- having an example of a board ext that carries state

Integration of the new dockViewR API is a thing for a separate PR. Here we're still on the old API.

PR #28, opened 2025-11-01
Update block UI

### LINKED ISSUE #9: Rework block UI
The current block UI is outdated.

### LINKED ISSUE #10: Include collapse state in block UI with serialized state
Currently, we do not keep track of block UI collapse state server-side. This should however be included in serialized board state and we therefore

1. have to track this server-side,
2. include it when serializing,
3. and apply when restoring.

@DivadNojnarg do you have ideas on how to tackle (1)?

This might be related to #9 and might be blocked by that so long as we have not found a way to go about (1).

DivadNojnarg:
Fixes #9, closes #10.

@nbenn Started to update the block ui. WIP

Note: there is a bit of duplication between blockr.dock and blockr.dag in term of color helpers.

### DIFF
--- a/R/plugin-block.R
+++ b/R/plugin-block.R
@@ -45,6 +45,7 @@ block_card_title <- function(block, id, info) {
         textInput(
           NS(id, "block_name_in"),
           "Block name",
+          value = block_name(block),
           updateOn = "blur"
         )
       )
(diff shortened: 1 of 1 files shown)


nbenn:
@DivadNojnarg given that blockr.dag will always depend on blockr.dock (while blockr.dock has no "hard" dependency on blockr.dag), we probably should move duplicated color/icon helpers to blockr.dock, export from there and use in blockr.dag.

nbenn:
@DivadNojnarg we now have in core `registry_id_from_block()` and `block_metadata()` which allows for a clean lookup of metadata for a block, i.e. `block_metadata(registry_id_from_block(blk))`.

Also, we have new block categories in core, so [`blk_color()`](https://github.com/BristolMyersSquibb/blockr.dag/blob/252b3a75b09a9a0c6914e4e698794b715549a264/R/utils.R#L19-L34) needs updating.

nbenn:
Now `blks_metadata()` also works with single blocks. Maybe this is all you need?

DivadNojnarg:
@nbenn Ready for review.

DivadNojnarg:
Only one thing was not implemented: the append block/remove block from the dropdown menu.

PR #29, opened 2025-11-01
proof of concept: Using the new icons and categories in the header.

christophsax:
Proof of concept related to https://github.com/BristolMyersSquibb/blockr.dag/pull/24 and https://github.com/BristolMyersSquibb/blockr.dag/pull/25. 

Since we need the icons and the colors here as well, we may have to move these functions to core, no? @nbenn?

It's gonna look nice once we have these icons everywhere:

image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,6 +22,7 @@ Imports:
     shiny,
     shinyjs,
     bslib,
+    bsicons,
     glue,
     dockViewR (>= 0.2.1),
     htmltools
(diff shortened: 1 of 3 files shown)


nbenn:
@christophsax have you tried circular nodes in the DAG with icons? Somehow I find the square icons in the DAG look a bit funny. But maybe this is just b/c I'm used to circular nodes?

christophsax:
Yes, I tried. Really, I find the iPhone 'squares' better but can give you a comparison. 

I would like to use the same thing in dag / select / title.

PR #41, opened 2025-11-13
vanilla busy indicator + button styling

DivadNojnarg:
@nbenn Restoring missing elements from old UI.

At a later point, we can use bs_theme for further customisation.


nbenn:
Thanks @DivadNojnarg!

PR #48, opened 2025-11-21
Simple panel layouting

### LINKED ISSUE #7: Helpers for `new_board_layout()`
Create some helper functions to generate board layouts via `new_board_layout()`.

nbenn:
Closes #7. Might move this over to dockViewR, depending on how generalizable it turns out to be. Maybe also a "partial" move of some utils. TBD.


DivadNojnarg (approved):
Thanks. We can check at a later point what should go to dockViewR: https://github.com/cynkra/dockViewR/tree/layout-helpers

PR #51, opened 2025-11-29
Style tweaks

christophsax:
Comprehensive set ot non-functional style tweaks that should make life with blockr more fun:

image

Each tweak has its own commit and a relatively detailed description.

The most consequential change are the bootstrap overviews. They make it easier for block authors to just use default shiny / bootstrap classes and get a look that work well in blockr. This includes a change of the default font size from 16 to 14, which leads to a more professional and modern appearance.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -19,6 +19,7 @@ Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.3
 Imports:
     blockr.core (>= 0.1.1),
+    bsicons,
     shiny,
     shinyjs,
     bslib,
--- a/R/board-ui.R
+++ b/R/board-ui.R
@@ -58,8 +58,8 @@ options_ui <- function(id, x, ...) {
   tagList(
     blockr_fab_dep(),
     tags$button(
-      class = "blockr-fab",
-      icon("gear"),
+      class = "blockr-edge-tab",
+      bsicons::bs_icon("chevron-left"),
       `data-bs-toggle` = "offcanvas",
       `data-bs-target` = paste0("#", offcanvas_id),
       `aria-controls` = offcanvas_id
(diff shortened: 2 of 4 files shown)


DivadNojnarg (commented):
Thanks @christophsax.

For now I am only providing a non technical review. I only looked at the UI result:

- The editable titles: I like the on-hover trigger. Can we close the popover when we click anywhere outside? Now we have to click again on the title to close the popover.
- For the FAB: I like the idea, it's not visible enough, maybe have a more curved form. In general, I would have liked to have the board options in the off-canvas triggered via this side button, but move the quick actions (save, browse, show code) to an easier to access space, ideally something that shows on hover and not on click. An example here: https://ionicframework.com/docs/api/fab#positioning, except that the trigger should not be click but hover.
- I have mixed feelings about the input container. It looks ok for more complex blocks with many inputs (ggplot) but too heavy for mono inputs blocks (like head). Also have a panel container, an input container, this adds too many borders. Currently, the input panels seems to have a relief. What happens, if instead we apply a shadow effect to the outer dockview panel and have a sort of negative shadow for the input panels so it appears more nested?

talpitoo:
> I'd check with @talpitoo what he thinks here.

- one option is to add background color https://getbootstrap.com/docs/5.3/utilities/background/#background-color to the input fields (which change depending on dark/light mode in the latest Bootstrap)
- At https://cynkra.com/contact/ and https://cynkra.com/ui-library/#input-fields we have a default background color which is different from the page/section default background (you can test it by toggling the dark/light mode in the footer)

...but since i am not familiar with this package, you can discard my comment as just an opinion.

christophsax:
@DivadNojnarg Let's continue discussion on side bar menu here: https://github.com/BristolMyersSquibb/blockr.dock/issues/53, it's not related to this PR.

nbenn (approved):
LGTM, thanks.

PR #56, opened 2025-12-01
Side bar window dressing

christophsax:
This is pure window dressing, since a proper solution requires plugins to think about quick actions and more complex UI and need more thoughts - for V2.

For now, this overrides a few things that come from the plugins to make it appear more congruent (such as removing the icon on the export button, which disrupts the alignment). It's pure CSS and harmless.

image

Comes along with a fix to the bootstrap overrides, which disrupted the file input element.


nbenn (approved):
LGTM, thanks!

PR #79, opened 2025-12-15
Actions V2

### LINKED ISSUE #66: Redo "actions"
Some issues:

- We currently have per action multiple places where the same code runs. This could be centralized such that we may have multiple triggers that are registered somehow with the single instance of an action.
- The current implementation is a mess w.r.t. scoping. This needs to improve.
- Currently we don't get unit test coverage for actions due to the mangling of expressions into functions. Do we really need to be able to create actions as modules *and* "inline"? If we have a single place where an action runs, all of this goes away.

Maybe an API like

```r
some_action <- function(trigger, board, update, ...,
                             domain = get_session()) {

  new_action(
    function(input, output, session) {
      # some code
    }
  )
}
```

would feel more natural?

nbenn:
Closes #66.


DivadNojnarg (approved):
@nbenn It's good to go. Just make sure the documentation is ready with the new API. I've seen outdated elements in the roxygen doc for the actions.

nbenn:
@DivadNojnarg i'll merge this in a couple of mins. Will break current main for blockr.dag unless we merge https://github.com/BristolMyersSquibb/blockr.dag/pull/83. So I'll merge that one too.

PR #80, opened 2026-01-06
Feature/navbar provider

christophsax:


christophsax:
Let's discuss this first, I guess I am a bit confused about plugins and extensions. The first are from blockr.core, the second from blockr.dock, right?

Instead of the navbar provider class (which is a special kind of dock extension) I would have preferred normal dock extensions. But they seem to always add their own panel? Also, we may not want to serialize it (arguably).

So perhaps we could extend the extension framework?

 ```r
  new_dock_extension(
    ...,
    panel = FALSE,      # Don't create dock panel
    serialize = FALSE   # Don't include in board serialization
  )
```

Previously, blockr.session was a plugin, not an extension, right? But since blockr.core does not know about the navbar, that doesn't work.

christophsax:
abandon in favor of #81

PR #81, opened 2026-01-07
Feature/navbar preserve board

christophsax:
This works well without plugin, good enough with manage session and nicely with the new manage project.

image


nbenn (approved):
LGTM, thanks!

PR #82, opened 2026-01-13
align CSS with Figma

christophsax:
This makes the CSS closer aligned with the Figma mockup, and with the new navbar.

In general, the input elements are more modern, and rounder.

Before:
image

After:
image

And here you can see it in action: https://blockr.cloud/app/dock/


nbenn (approved):
LGTM, thanks.

PR #85, opened 2026-01-24
Feature/sidebar panels

christophsax:
I reworked the block / link / stack / panel selection in a new sidebar. The sidebar can be used similarly to a modal and we should also be able to use it from blockr.ai or from other blocks. You can play with it here: 

https://blockr.cloud/app/dock-sidebar


christophsax:
closing, in favor of #87

PR #87, opened 2026-02-01
Feat/sidebar s3 dispatch

christophsax:
Next try:
- Sidebar is JS only
- Content provided by S3 method

TBD


karoliskoncevicius:
@christophsax Mike said this -  https://github.com/BristolMyersSquibb/blockr.dock/issues/94 - is probably related, so linking it here

christophsax:
will be re-implemented as part of blockr.ui, will push an updated prototype

PR #98, opened 2026-03-18
sparkle UI

christophsax:


christophsax:
updated PR, @nbenn

christophsax:
@nbenn could please have a look at these, thanks!

nbenn:
The plugin contract is fine. blockr.core's `ctrl_block_ui(id, x)` and blockr.ai's `ai_ctrl_ui(id, x)` use the same signature, and the dock board's call site in `R/block-ui.R:40` invokes it as `ctrl(NS(blk_srv_id, "ctrl_block"), blk)`. So nothing crashes — blockr.core's plain text-input + Submit UI renders correctly inside the panel.

## Issues I found by running the board with `blockr.core::ctrl_block()`

**1. The wrapper labels and brands the plugin as AI, regardless of what plugin is in use.** In `R/plugin-block.R`:
- `ctrl_header_icon` (line 369) hard-codes a sparkle SVG, `class="blockr-sparkle-btn"`, and `title="Toggle AI assistant"`
- `ctrl_collapsible_section` (line 422) hard-codes the section header text `"AI Assistant"`

Result with blockr.core's plugin: a sparkle button labelled "Toggle AI assistant" opens a section labelled "AI ASSISTANT" containing… plain text inputs and a Submit button.

**2. The dock board doesn't gate on `block_supports_external_ctrl(x)`.** Compare to blockr.core's own `block_ui.board`, which only renders the ctrl slot when `block_supports_external_ctrl(x) && !is.null(ctrl)`. The PR drops that check (`R/block-ui.R:40`):

```r
ctrl_tag <- if (!is.null(ctrl)) ctrl(NS(blk_srv_id, "ctrl_block"), blk)
```

So every block — even ones with `external_ctrl = FALSE` — gets the sparkle button. Clicking it on a `head_block` opens an "AI ASSISTANT" section with no contents, since `ctrl_block_ui` returns an empty `tagList()` for blocks without external control. (`is.null(ctrl_ui)` in `ctrl_collapsible_section` doesn't catch this — an empty tagList isn't NULL.)

This problem also exists with blockr.ai's plugin, but is masked because `ai_ctrl_ui` is typically used with chat-enabled blocks.

## Proposed direction

Treat the control plugin like Input/Output: render a toggle button next to those, expanding into a panel that mirrors the inputs accordion. Show the button only when the plugin is provided **and** `block_supports_external_ctrl(blk)` is true. The default appearance is a generic "Control" button; specific variants (e.g. blockr.ai's chat) can customise the label/icon/styling.

### Customisation seam

The cleanest hook is to attach metadata as **attributes on the plugin's UI function itself**. Function attributes are readable without invoking the function, persist through `plugin_ui()` (which just does `x[["ui"]]`), survive package install, and don't require any cross-package S3 dispatch. They also avoid the awkwardness of subclassing the plugin object — `plugin_id()` is hardcoded to `class(x)[1L]` and `names.plugins()` derives from it, so adding a subclass like `c("ai_ctrl_block", "ctrl_block", "plugin")` would break the `"ctrl_block" %in% names(plugins)` lookup.

blockr.ai opts in like this:

```r
ai_ctrl_ui <- structure(
  function(id, x) {
    # ...existing body...
  },
  ctrl_label = "AI",
  ctrl_icon  = sparkle_svg(),
  ctrl_class = "blockr-sparkle-btn"
)

ai_ctrl_block <- function() {
  blockr.core::ctrl_block(server = ai_ctrl_server, ui = ai_ctrl_ui)
}
```

blockr.dock provides accessors with sensible defaults (so blockr.core's vanilla `ctrl_block_ui` — which carries no attributes — falls through to the generic look automatically, no changes needed in core):

```r
ctrl_btn_label <- function(fn) coal(attr(fn, "ctrl_label"), "Control")
ctrl_btn_icon  <- function(fn) coal(attr(fn, "ctrl_icon"),  bsicons::bs_icon("sliders"))
ctrl_btn_class <- function(fn) attr(fn, "ctrl_class")
```

This keeps blockr.ai independent of blockr.dock (no soft dep needed), keeps blockr.core untouched, and gives plugin authors a single, documented place to control how their plugin presents itself.

## Concrete changes for this PR

1. In `block_card`: only construct `ctrl_tag` when `block_supports_external_ctrl(blk) && !is.null(ctrl)`. If either is false, skip the toggle button entirely.
2. Replace `ctrl_header_icon` with a generic toggle styled like the existing Input/Output toggles, reading label/icon/class from the UI function's attributes via the accessors above.
3. Move the AI-specific sparkle SVG, CSS, and animations into blockr.ai, attached to `ai_ctrl_ui` via the attribute convention.
4. Drop the hardcoded "AI Assistant" header from `ctrl_collapsible_section`; the section just renders the plugin UI inside a panel that mirrors the existing inputs/outputs accordion.

nbenn:
@christophsax do we merge this as is? Anything left to do here?

nbenn (commented):
Lgtm, pending any further design tweaks.

christophsax:
Shall we rebase and then I am applying the design tweaks? 

Or you merge and I will do a design PR afterwards? (my preference, but not important)

nbenn:
@christophsax let's add related stuff here. Feel free to rebase. Will merge as soon as you tell me it is ready.

christophsax:
ready to go, @nbenn

nbenn (approved):
Lgtm, thanks!

PR #104, opened 2026-04-03
hideable, cleaned up header

christophsax:
second PR, this implements more minimal visibility triggers, and a new possibility to hide the header.


christophsax:
Abandoned, no header hiding.

PR #105, opened 2026-04-08
Feat/sparkle UI hide

christophsax:
First commit is functional, second is CSS only.

- hide header
- equally sized header buttons that look nice
- get tooltip from blockr.ai


christophsax:
Abandoning, hide header is canceled. 

I can do a PR with the two style tweaks:

- equally sized header buttons that look nice
- get tooltip from blockr.ai

PR #103, opened 2026-04-01
Feat pages

DivadNojnarg:
TODO:

- [x] Check ser/deser
- [x] Simplify code
- [x] Rename workspace to something better like `page` (I don't like neither). `view` was the perfect semantic element from a web POV, however it is used by dockViewer. Maybe we can dig to see if we can solve this ...
- [x] Fix dynamic ws inserting offset issue. Due to bootstrap 5 conflicts:
Screenshot 2026-04-01 at 09 20 26

## Architecture summary

See documents used at https://github.com/cynkra/blockr.design/tree/main/open/global-tabs

image

```mermaid
graph TD
    A["serve(board, 'my_board')"] -->|S3 dispatch| B["blockr_app_ui.dock_board()"]
    A -->|S3 dispatch| C["blockr_app_server.dock_board()"]

    %% ── UI BUILD ──
    B --> B1["board_ui()"]
    B1 --> B2["Blocks & extensions<br/>rendered in offcanvas"]
    B1 --> B3["workspace_nav_ui()<br/>dropdown tabs"]
    B1 --> B4["dock_outputs_ui()<br/>#ws_container"]

    %% ── SERVER INIT ──
    C --> C1["board_server(callback = board_server_callback)"]
    C1 --> C2["board_server_callback()"]
    C2 --> D{"workspaces?"}

    %% ── MULTI-WORKSPACE PATH ──
    D -->|YES| E["new_dock_manager()<br/>docks + active_dock + next_id"]
    E --> E1["init_workspace_docks()"]
    E1 -->|"for each ws"| E2["insertUI → manage_dock()"]
    E2 --> E3["update_active_dock()<br/>seed active_dock from active ws"]

    E3 --> F["Register observers"]
    F --> F1["switch_workspace_observer"]
    F --> F2["add_ws_observer"]
    F --> F3["remove_ws_observer"]
    F --> F4["rename_ws_observer"]

    %% ── SINGLE-DOCK PATH ──
    D -->|NO| G["manage_dock('dock_main')"]

    %% ── EXTENSIONS + ACTIONS ──
    E3 --> H["extension_server(dock = active_dock)"]
    G --> H
    H --> I["register_actions()"]

    %% ── manage_dock INTERNALS ──
    E2 --> MD["manage_dock()"]
    MD --> MD1["set_dock_view_output()"]
    MD1 --> MD2["restore_dock(layout)"]
    MD2 --> MD3["show_block_panel / show_ext_panel<br/>move DOM from offcanvas → dock panel"]
[...]

PR #106, opened 2026-04-20
fix: action tests broken by reactiveVal scoping in testServer

DivadNojnarg:
@nbenn for the CRAN issue. To merge before #103 

- reactiveVal/reactiveValues created inside the testServer app function live in a different reactive domain than the module's closure. The action handler sets the reactiveVal but the test expression reads a different copy. Fix: create them outside testServer.

```deps
BristolMyersSquibb/blockr.dag#111
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -39,6 +39,8 @@ Suggests:
     methods,
     withr,
     shinytest2
+Remotes:
+    BristolMyersSquibb/blockr.core
 Config/testthat/edition: 3
 Collate:
     'action-class.R'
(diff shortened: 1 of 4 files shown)


DivadNojnarg:
CICD fails because blockr.dag isn't merged yet: https://github.com/BristolMyersSquibb/blockr.dag/pull/111

nbenn:
@DivadNojnarg thanks for the debugging & fixing here. Looks good. I'm trying to unblock the ci failure and then we can start merging this stuff!

nbenn (approved):
Lgtm, thanks!

PR #107, opened 2026-04-28
Cover #102 stack/link modal crash; fix in blockr.core

nbenn:
The actual fix for #102 has moved upstream to [blockr.core#170](https://github.com/BristolMyersSquibb/blockr.core/pull/170): `block_metadata()` now returns an empty data.frame for an empty `blocks` input rather than `NULL`, so the `cbind()` in `blks_metadata()` stays a data.frame and the create-stack/link modals open cleanly on a fresh board.

This PR is now just the dock-side regression test plus a temporary `Remotes:` pin to the upstream branch so dock CI exercises the fix. Once #170 lands the pin is dropped.

Closes #102

PR #108, opened 2026-04-28
fix: dock app shinytest2 fixture and refresh README

DivadNojnarg:
@nbenn Will fix the shinytest2 issue you have in #107

The shinytest2 test in test-utils-serve.R hardcodes `my_board-...` selectors and relied on an empty board. Commit 7e6eaea ("better readme") rewrote inst/examples/dock/app.R with a richer board (dataset + scatter, preset link, no `my_board` id) so the test could no longer find its selectors during devtools::check().

Add a dedicated minimal fixture at inst/examples/empty/app.R and point the test at it, leaving the rich `dock` example untouched for the README showcase.

Also rebuild README/figures so the views snippet uses the new dock_view(active = TRUE) syntax instead of the removed `dock_layouts(..., active = ...)` parameter.


README.md
```diff
@@ -102,11 +102,10 @@ board <- new_dock_board(
     head_1 = new_head_block()
   ),
   links = new_link("dataset_1", "head_1"),
-  layout = dock_layouts(
```
DivadNojnarg, line 105: remove leftover

tests/testthat/test-utils-serve.R
```diff
@@ -33,7 +33,7 @@ test_that("dock app", {
   skip_on_cran()
 
   app <- shinytest2::AppDriver$new(
-    system.file("examples", "dock", "app.R", package = "blockr.dock"),
+    system.file("examples", "empty", "app.R", package = "blockr.dock"),
```
DivadNojnarg, line 36: now refer to correct app for the test

nbenn (approved):
Lgtm, thanks.

PR #109, opened 2026-04-28
docs: surface views API in pkgdown reference

DivadNojnarg:
Fixes pkgdown failure.

Add a "Dock views" section to the reference index pulling in the multi-view API from man/view.Rd (new_dock_layouts, dock_view, active_view and friends), which was previously uncategorised and caused `pkgdown::check_pkgdown()` to error.

Also fix a typo in the Utilities section description ("dependen" -> "dependent").


nbenn (approved):
Thanks

PR #115, opened 2026-04-29
Collapse dock_layout/dock_grid; add panels(), group(), sizes

### LINKED ISSUE #114: Rename default_view_grid → default_layout, unexport default_grid
## Problem

The default in `new_dock_board()` reads:

```r
layout = dock_layouts(Page = default_view_grid(blocks, extensions))
```

Two issues:

1. `default_view_grid()` is **not exported**, even though it appears in an exported function's default formals. Users who want to extend or replicate the default cannot reach it.
2. `default_view_grid` mixes the conceptual term ("view" = a page) with the technical term ("grid" = a layout's panel arrangement). In our model a view *has* a layout; the function is producing a default layout spec for a view, so `default_layout` is the accurate name.

There is also a parallel `default_grid()` that emits panel IDs (rather than block/extension names). It is consumed only by `create_dock_layout()`, which already does names→panel-ID translation internally. Keeping two flavours of `default_grid` exported invites confusion.

## Proposed change

- Rename `default_view_grid()` → `default_layout()` and **export** it. This is the names-form default, the one that goes inside a view spec consumed by `dock_layouts()`.
- Rename `default_grid()` → `default_panel_grid()` and **unexport** it. Panel-ID form is an internal concern of `create_dock_layout()`.
- Update `new_dock_board()`'s default to `dock_layouts(Page = default_layout(blocks, extensions))`.
- Update `vignettes/layouts.Rmd` to (a) use the new name and (b) briefly explain the view/layout distinction (a view is a page; its layout is the panel arrangement inside it).
- Regenerate Rd files.

No behaviour change — purely a rename + export hygiene fix.

## Rationale

- "view" is the conceptual / page-level term (used in `dock_view()`, `active_view()`, `board_views()`, `view_ids()`).
- "layout" is the technical structure inside a view (`dock_layout`, `create_dock_layout()`, `dock_layouts()` keyed by view name).
- A function producing a default *layout spec* should be `default_layout`, not `default_view_grid`.

nbenn:
Closes #114

The trigger was the small rename + export hygiene fix in #114. The audit surfaced bigger issues — overlapping concepts, a latent staleness bug from panels being stored per-view, and a serialized format that tracked dockview's internal tree shape — so the scope grew into a layout-API refactor.

## Public layout API

| | What | Returns |
|---|---|---|
| `dock_layout(..., orientation, active, sizes)` | per-view arrangement | `dock_layout` |
| `panels(..., active = NULL)` | tabbed leaf with explicit open tab | `dock_panels` node |
| `group(..., sizes = NULL)` | branch with explicit child ratios | `dock_group` node |
| `default_layout(blocks, extensions)` | default two-row arrangement | `dock_layout` |
| `is_dock_layout(x)` | predicate | boolean |
| `as_dock_layout(x)` | coercion (board, list, dock_layout) | `dock_layout` |
| `validate_dock_layout(x, blocks)` | validator | input (throws on error) |
| `is_dock_layouts(x)` | predicate | boolean |
| `as_dock_layouts(x)` | coercion (dock_layouts, dock_layout) | `dock_layouts` |
| `validate_dock_layouts(x)` | validator | input (throws on error) |
| `active_view(x)` / `active_view(x) <-` | name of the active view (S3 on `dock_layouts` and `dock_board`) | string |
| `active_layout(brd)` / `active_layout(brd) <-` | the active resolved layout on a board | `dock_layout` |
| `board_layouts(brd)` / `board_layouts(brd) <-` | the full collection on a board | `dock_layouts` |
| `layout_to_json(x, ...)` | layout → JSON wire string | string |
| `layout_from_json(x, blocks, extensions)` | JSON / list → layout (resolves + validates when given board components) | `dock_layout` |
| `layout_panel_ids(layout)` | canonical panel IDs referenced by a layout | character |
| `panel_obj_ids(ids)` | strip `block_panel-` / `ext_panel-` prefixes to bare IDs | character |

The dockview wire shape (`grid + panels`) is not a public type; `dockview_payload(layout, blocks, extensions)` materialises it on demand at the dockview boundary.

## Why per-view panels are gone

Each `dock_layout` used to store its own resolved `panels` map. Panels are derivable from the board's blocks and extensions, so the storage was redundant — and stale: when a block was renamed, only the active view's live proxy was updated; inactive views kept the old title in their stored `panels` until next interaction.

Per-view storage is now `{grid: tree, activeGroup: string}` only.

## Wire format decoupled from dockview

The serialized form is our own spec, not dockview's tree. The top object is itself a branch: it carries `orientation`, `children`, and an optional `sizes`. A child is a bare string (single-panel leaf), a `{panels, active?}` object (tabbed leaf), or a `{children, sizes?}` object (nested branch). Bare arrays never appear as node values, so the three forms are unambiguous. Sizes are ratios summing to 1 (auto-normalised from dockview's pixel sizes); even splits omit the field; default open tabs omit `active`. Dockview's internal tags / per-branch sizes / leaf ids / activeGroup aren't stored — they're regenerated by `wire_to_grid()` on load.

`layout_to_json()` / `layout_from_json()` are the canonical reader/writer for this format (so downstream tooling — e.g. blockr.assistant — needs no second implementation). A single-panel example:

```json
{"orientation":"horizontal","children":["a",{"panels":["b","c"],"active":"c"}],"sizes":[0.3,0.7]}
```

Legacy payloads keep loading via a shape-discriminated reader inside `blockr_deser.dock_layout` (presence of a `grid` field flags the old shape). Full version-based routing is queued behind blockr.core forwarding `...` through `blockr_deser.list` (#153, blockr.core#186).

## Breaking changes (pre-1.0, no compat shims)

- `dock_layout(...)` is now the per-view *constructor*. The same name on main was a board accessor — that role moves to `active_layout(brd)`. The stored shape also changes: no `panels` field.
- `new_dock_layout()` is unexported; use `dock_layout()`.
- Board accessors renamed — `dock_layout(brd)` → `active_layout(brd)`, `board_views(brd)` → `board_layouts(brd)`, `dock_layouts(brd) <-` → `board_layouts(brd) <-`.
- `new_dock_board()` parameter `layout` → `layouts`. Default is `default_layout(blocks, extensions)`.
- `dock_view()` removed; use `dock_layout(active = TRUE)`.
- `default_grid()` removed. Use `as_block_panel_id()` / `as_ext_panel_id()` if you need panel-ID strings.
- `create_dock_layout()` unexported (now internal `resolve_dock_layout()`).
- `dock_layouts()` constructor unexported. `is_dock_layouts()`, `as_dock_layouts()`, `validate_dock_layouts()` remain exported.
- `active_view()` / `active_view<-()` are now S3 generics (added a `dock_board` method).
- `view_ids()` → `layout_ids()` (internal); `view_can_crud()` → `views_can_crud()` (internal).
- Serialised layout payloads use our own flattened wire spec; pre-PR payloads still load via the legacy reader.

## Migration

| Before | After |
|---|---|
| `new_dock_board(layout = ...)` | `new_dock_board(layouts = ...)` |
| `dock_layout(brd)` | `active_layout(brd)` |
| `dock_layout(brd) <- ly` | `active_layout(brd) <- ly` |
| `board_views(brd)` | `board_layouts(brd)` |
| `dock_layouts(brd) <- ly` | `board_layouts(brd) <- ly` |
| `dock_view(...)` | `dock_layout(..., active = TRUE)` |
| `default_grid(...)` | gone; use `as_block_panel_id()` / `as_ext_panel_id()` for panel-ID strings |
| `create_dock_layout(...)` | gone (was a niche internal-style entry point) |
| `new_dock_layout(grid, panels, ...)` | `dock_layout(...)` (different signature — see API table) |

Follow-ups: #138 (`layout_preview()` HTML helper), #153 (version-based ser/des routing), blockr.core#186 (forward `...` in `blockr_deser.list`).


nbenn:
## Proposal: flatten the serialized layout wire shape

Context: building the LLM layout tools in blockr.assistant#19 (`list_views` / `add_view` / `modify_view`), the model reads a layout, edits the structure, and writes it back. The serialized shape is what it manipulates, so its ergonomics matter — and since we own the format, a couple of tweaks would make it both shorter and harder to get wrong.

### Current persisted shape

```json
{
  "orientation": "horizontal",
  "root": {
    "children": [
      {"children": [
        {"panels": ["data"]},
        {"children": [{"panels": ["head"]}, {"panels": ["scatter"]}]}
      ]},
      {"panels": ["assistant_extension"]}
    ],
    "sizes": [0.7, 0.3]
  }
}
```

Already unambiguous (every node tagged `panels` ⇒ leaf or `children` ⇒ branch), which is the important property. Two things make it heavier than it needs to be:

1. The `root` wrapper duplicates a level — the top is always a single node, and that node is almost always a branch.
2. `{"panels": ["data"]}` for a single-panel leaf is the overwhelmingly common case and the most verbose way to write "panel data".

### Proposed shape

```json
{
  "orientation": "horizontal",
  "children": [
    {"children": ["data", {"children": ["head", "scatter"]}]},
    "assistant_extension"
  ],
  "sizes": [0.7, 0.3]
}
```

Two changes:

- **Drop `root`.** Hoist `children` + `orientation` (+ optional `sizes`) to the top object. The top level becomes just a branch that additionally carries `orientation`. Inner branches are the same minus `orientation` (they alternate by depth).
- **Bare string = single-panel leaf.** `"data"` is sugar for `{"panels": ["data"]}`. Multi-panel / explicit-active leaves stay `{"panels": [...], "active": "..."}`.

### The invariant that keeps it unambiguous

**Bare arrays never appear as node values** — arrays only ever live inside a `panels` or `children` key. So the three node forms are syntactically distinct:

| Node | Form |
|---|---|
| Single-panel leaf | bare string `"data"` |
| Tab leaf | `{"panels": [...], "active"?: "..."}` |
| Branch | `{"children": [...], "sizes"?: [...]}` |

No `["a","b"]`-means-either problem (that bit us in assistant#19 — an even-split branch round-tripped into a tab leaf because both serialized to a bare array).

### Notes / your call

- This touches persistence — saved boards carry the current shape. The `blockr_deser.dock_layout` reader already shape-discriminates (the `grid`-field legacy check); adding a `root`-presence check to route old → new is the same pattern. Or keep persistence on the current shape and expose only the flattened form through the new serialize/deserialize API (proposed in the next comment) — but I'd lean toward one shape everywhere.
- `activeGroup` stays dropped (already the case) — dockview-internal focus state, regenerated on load, nothing the model should set.
- The bare-string sugar is optional. If you'd rather keep leaves uniformly `{"panels": [...]}` for a simpler `wire_to_grid` walk, dropping `root` alone still gets most of the win.

nbenn:
## API request: serialize/deserialize/inspect helpers so the assistant duplicates nothing

Right now blockr.assistant#19 carries its own copy of the layout serialization (`layout_to_spec` / `layout_from_spec` + recursive node walkers + a bare-id panel extractor — ~300 lines). That's the second implementation of logic dock already owns (`grid_to_wire` / `wire_to_grid`, `layout_panel_ids`, `resolve_dock_layout`, `validate_dock_layout`), and the two can drift — a dry-run probe could green-light something the real flush rejects. Everything the assistant does maps onto dock concepts; it just needs them as a public surface.

### What the assistant does, and what it should call instead

| Tool | Needs | Dock has it? |
|---|---|---|
| `list_views` | `dock_layout` → JSON | internal `grid_to_wire` (under `blockr_ser` envelope) |
| `add_view` / `modify_view` | JSON → `dock_layout` | internal `wire_to_grid` (under `blockr_deser`) |
| `validate_layout` (dry run) | parse + resolve bare IDs + validate + panel-ref check | `resolve_dock_layout` (internal), `validate_dock_layout` (exported) |
| `summarise_view` | referenced panel/object IDs of a layout | `layout_panel_ids` (`@noRd`) |

### Requested exports

1. **`layout_to_json(layout)`** → JSON string (or the list ready for `jsonlite::toJSON`). Whatever shape #115 settles on (see the flatten proposal above), this is the single canonical writer. Replaces `layout_to_spec`.

2. **`layout_from_json(x, blocks = NULL, extensions = NULL)`** → `dock_layout`. Accepts a JSON string or an already-parsed list (matching the `as_dock_layout.list` style). When `blocks` / `extensions` are supplied, it resolves bare IDs to canonical panel IDs and runs `validate_dock_layout` — i.e. the `resolve_dock_layout` + `validate_dock_layout` combo, behind one call. Throws the existing classed errors. Replaces `parse_layout_json` + `layout_from_spec` + the walkers, and makes `validate_layout` a literal `tryCatch(layout_from_json(json, blocks, exts), ...)` — no separate validation path to drift.

3. **`layout_panel_ids(layout)`** exported (it's already written, just `@noRd`). The assistant needs it for the dry-run cross-check and the per-view panel summary. If it returns canonical (`block_panel-…`) IDs, a paired `panel_obj_ids()` (also already internal) or a documented note on stripping would let me map back to bare block/extension IDs — whichever you prefer to expose.

With those three, `R/views-layout.R` in the assistant deletes entirely; the tools become thin wrappers (`jsonlite` + a dock call + a trivial "do these IDs exist on the board" set check, which is legitimately assistant-side state).

### Not requesting

- `resolve_dock_layout` on its own — folding resolution into `layout_from_json(blocks=, extensions=)` is cleaner than exposing it raw, unless you'd rather keep them separate.
- Anything that changes the apply/augment path. The assistant only touches read-side serialization and a pre-stage dry run.

No rush and no scope pressure from my side — the assistant PR works today with its local copy; this is purely about not maintaining two implementations of your format. I'll delete the duplication the moment the exports land.

nbenn:
Both done in `b9e39b3`.

**Flatten (A):** dropped `root` (hoisted `children`/`sizes`/`orientation` to the top object) and added bare-string single-panel leaves. A single-panel board now serializes to e.g.

```json
{"orientation":"horizontal","children":["a",{"panels":["b","c"],"active":"c"}],"sizes":[0.3,0.7]}
```

The all-single-panel-children case collapses to an atomic vector under `fromJSON(simplifyVector = TRUE)`; `wire_to_grid()` handles that (iterates via `Map`), and a real `toJSON`/`fromJSON` round-trip test guards it — our prior tests only did R-list round-trips, which is exactly why the assistant#19 boxing bug could hide.

**Exports (B):** `layout_to_json()` / `layout_from_json()` are the canonical reader/writer. `layout_from_json(blocks =, extensions =)` folds resolve + validate into the parse, so `validate_layout` becomes `tryCatch(layout_from_json(json, blocks, exts), ...)` with no separate path to drift. `layout_panel_ids()` and `panel_obj_ids()` are now exported for the dry-run cross-check and the bare↔canonical mapping. `R/views-layout.R` in the assistant should delete cleanly against these.

Producer-version routing (instead of the current shape discriminator) still tracked in #153 / blockr.core#186.

PR #117, opened 2026-05-03
fix: guard rename observer against missing dock panels

### LINKED ISSUE #116: Rename observer crashes with 'argument is of length zero' when a block has no dock panel
## Summary

Renaming a block raises `Error in if (new_name == old_title) ...: argument is of length zero` whenever the modified block (or any block currently marked modified at the same time) has **no dock panel** — for example because it is parked in the offcanvas, not placed in any `layout` view.

## Root cause

In `R/board-server.R` (`manage_dock`, around the `update()$blocks$mod` observer):

```r
old_title <- get_dock_panel(blk_panel_id, dock)$title

if (new_name == old_title) {
  next
}
```

`get_dock_panel(id, proxy)` returns `NULL` when no panel matches `id`, so `old_title` becomes `NULL`. The comparison `NULL == "something"` returns `logical(0)`, and `if (logical(0))` errors out — aborting the loop and the rename.

## Reproducible example

`inst/examples/rename-offcanvas-crash/app.R`:

```r
library(blockr.core)
library(blockr.dock)

board <- new_dock_board(
  blocks = c(
    a = new_dataset_block("iris"),
    b = new_dataset_block("mtcars")
  ),
  extensions = list(edit = new_edit_board_extension()),
  layout = list("edit", list("a"))   # b is parked in the offcanvas
)

serve(board, "repro_rename_offcanvas")
```

Steps:
1. Open the app — `b` is in the offcanvas (it is not in `layout`).
2. Open the edit extension and rename `b`.
3. The Shiny session errors with `argument is of length zero`.

The same crash also fires during dashboard restore: many blocks are reported as `mod` at once and the observer iterates over all of them, so renaming *any* block crashes as soon as one panel-less entry is reached.

## Suggested fix

Guard the comparison against length-zero inputs and switch to `identical()`:

```r
old_title <- get_dock_panel(blk_panel_id, dock)$title

if (length(old_title) == 0L || length(new_name) == 0L ||
    identical(new_name, old_title)) {
  next
}
```

Branch with the fix: `fix/rename-observer-length-zero` (commit cbb12d9).

christophsax:
Fixes #116.

## Problem

Renaming a block crashes the per-view rename observer with `Error in if (new_name == old_title): argument is of length zero` whenever the renamed block is not present in *this* view's dock — most commonly because:

- the block is parked in the offcanvas (no panel anywhere), or
- the board has multiple views and the block lives in only some of them.

## Root cause

`R/board-server.R` (`manage_dock`, the `update()$blocks$mod` observer) calls

```r
old_title <- get_dock_panel(blk_panel_id, dock)$title
if (new_name == old_title) next
```

`get_dock_panel()` returns `NULL` when no panel matches the id, so `old_title` is `NULL` and `NULL == "..."` is `logical(0)`, which crashes the `if`.

## Fix

Short-circuit on length zero before the equality check, and use `identical()` for the comparison. No-op when the block has a panel and the title actually changed.

```r
if (length(old_title) == 0L || length(new_name) == 0L ||
    identical(new_name, old_title)) {
  next
}
```

### DIFF
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -473,7 +473,12 @@ manage_dock <- function(
 
           old_title <- get_dock_panel(blk_panel_id, dock)$title
 
-          if (new_name == old_title) {
+          # Skip when there's no panel for this block (e.g. card still in
+          # the offcanvas) or when the name didn't actually change. Without
+          # the length guard, comparing against a NULL/empty old_title
+          # produces logical(0) and crashes the `if`.
+          if (length(old_title) == 0L || length(new_name) == 0L ||
+              identical(new_name, old_title)) {
             next
           }
 


nbenn:
Superseded by #307, closing.

As flagged in the #116 thread, this patch guards the `new_name == old_title` line, but #202 has since rewritten that line to `identical(new_name, old_title)`, so the diff no longer applies to `main`. Making the comparison `NULL`-safe also moved the failure rather than removing it: for a block absent from a given view the observer now falls through to `set_panel_title` and dockView throws client-side (`Cannot read properties of undefined (reading 'api')`) instead of raising `argument is of length zero`.

#307 fixes the root cause you identified — the per-view rename observer touching a panel the view doesn't hold — by gating the title push on `live_panels()`, the authoritative server-side membership set the module already maintains, so a view only relabels panels it actually holds. Thanks for the diagnosis and both reproducers on #116; the multi-view case is now a regression test.

PR #121, opened 2026-05-07
Phase 2: adopt blockr.ui sidebar primitive

DivadNojnarg:
## Summary

Adopts the sidebar primitive from \`blockr.ui\` (PR \`BristolMyersSquibb/blockr.ui#2\`) for board-mutation flows and the navbar settings gear.

- Six action handlers (\`add_block\`, \`append_block\`, \`prepend_block\`, \`add_link\`, \`add_stack\`, \`edit_stack\`) drop \`showModal\`/\`removeModal\` and now use \`blockr.ui::show_sidebar()\` + \`blockr.ui::keep_or_hide_sidebar()\`. On confirm, a pinned panel re-renders a fresh form for chaining; an unpinned panel closes — same code path for all six.
- Modal builders (\`block_modal\`, \`link_modal\`, \`stack_modal\`) become \`*_sidebar_body()\` (return plain \`tagList()\`); file renamed \`R/action-modal.R\` → \`R/action-sidebar.R\`.
- The navbar gear becomes a regular \`actionButton(NS(id, "settings_btn"))\`; the Bootstrap \`options_offcanvas\` markup is removed.
- A new \`settings_observer()\` opens the sidebar with \`settings_body(id, board$board)\` on click. \`settings_body()\` uses \`blockr.core::blockr_app_options(x)\` (not bare \`board_options(x)\`) so the accordion includes block-contributed sections — same set \`serve()\` passes to \`board_ui()\`.
- \`board_ui.dock_board()\` mounts \`blockr.ui::sidebar_ui("main_sidebar")\` at top level (unwrapped by \`NS()\`); deep action handlers reach it via \`rootScope\` walking inside \`show_sidebar()\` — no session-proxy plumbing.

## Test plan

- [x] \`devtools::test()\` — 372 / 372 pass
- [x] \`devtools::check()\` — 0 errors / 0 warnings / 1 NOTE (timestamp only)
- [x] Headless chromote run against \`inst/examples/single-page\`:
  - pin / unpin / Esc / outside-click / X-button all behave per the sidebar spec
  - settings sidebar matches \`main\`'s offcanvas: two accordion sections (\`Board options\`, \`Table options\`) with \`board_name\` + \`n_rows\` + \`page_size\` + \`filter_rows\`
  - confirm while pinned re-renders a fresh form
- [ ] Manual smoke test of the \`dock\` + \`dag\` example (right-click append node, "+" toolbar button, settings gear)

## Sequencing

\`Remotes: BristolMyersSquibb/blockr.ui@phase1-sidebar-primitive\` is pinned to that PR's branch until it merges. Once \`blockr.ui#2\` lands, this PR's \`Remotes:\` line drops the \`@phase1-sidebar-primitive\` suffix.

## Out of scope

The block-browser UX (collapsible cards by category) is a separate follow-up; this PR ships the existing modal bodies into the sidebar unchanged so the diff stays mechanical.

### DIFF
--- a/R/action-link.R
+++ b/R/action-link.R
@@ -7,7 +7,13 @@ add_link_action <- function(trigger, board, update, ...) {
         trigger(),
         {
           body <- link_sidebar_body(session$ns, board$board, trigger())
-          if (is.null(body)) return()
+          if (is.null(body)) {
+            notify(
+              "No inputs are currently available.",
+              type = "warning"
+            )
+            return()
+          }
 
           blockr.ui::show_sidebar(
             "main_sidebar",
@@ -91,10 +97,30 @@ add_link_action <- function(trigger, board, update, ...) {
             list(links = list(add = new_lnk))
           )
 
-          blockr.ui::keep_or_hide_sidebar(
-            "main_sidebar",
-            title = "Create new link",
-            ui = link_sidebar_body(session$ns, board$board, trigger())
+          # Pinned sidebar: rebuild the form for another link, or close
+          # silently if the link we just added consumed the last input.
+          # `update()` only sets a reactiveVal; the board state is mutated
+          # on the next reactive flush, so defer the availability check
+          # until after the flush — otherwise `board$board` is still the
+          # pre-link snapshot. `onFlushed` runs outside a reactive context,
+          # so `isolate()` is needed to read `board$board`.
+          src_id <- trigger()
+          session$onFlushed(
+            function() {
+              isolate({
+                body <- link_sidebar_body(session$ns, board$board, src_id)
+                if (is.null(body)) {
+                  blockr.ui::hide_sidebar("main_sidebar")
+                } else {
+                  blockr.ui::keep_or_hide_sidebar(
+                    "main_sidebar",
+                    title = "Create new link",
+                    ui = body
+                  )
+                }
+              })
+            },
+            once = TRUE
           )
         }
       )
--- a/R/action-link.R
+++ b/R/action-link.R
@@ -3,6 +3,8 @@ add_link_action <- function(trigger, board, update, ...) {
   new_action(
     function(input, output, session) {
 
+      sidebar_id <- NS(isolate(board$board_id), "actions_sidebar")
+
       observeEvent(
         trigger(),
         {
@@ -16,7 +18,7 @@ add_link_action <- function(trigger, board, update, ...) {
           }
 
           blockr.ui::show_sidebar(
-            "actions_sidebar",
+            sidebar_id,
             title = "Create new link",
             ui = body
           )
@@ -101,7 +103,7 @@ add_link_action <- function(trigger, board, update, ...) {
           # silently if the link we just added consumed the last input.
           # `update()` only sets a reactiveVal; the board state is mutated
           # on the next reactive flush, so defer the availability check
-          # until after the flush — otherwise `board$board` is still the
+          # until after the flush; otherwise `board$board` is still the
           # pre-link snapshot. `onFlushed` runs outside a reactive context,
           # so `isolate()` is needed to read `board$board`.
           src_id <- trigger()
@@ -110,10 +112,10 @@ add_link_action <- function(trigger, board, update, ...) {
               isolate({
                 body <- link_sidebar_body(session$ns, board$board, src_id)
                 if (is.null(body)) {
-                  blockr.ui::hide_sidebar("actions_sidebar")
+                  blockr.ui::hide_sidebar(sidebar_id)
                 } else {
                   blockr.ui::keep_or_hide_sidebar(
-                    "actions_sidebar",
+                    sidebar_id,
                     title = "Create new link",
                     ui = body
                   )
--- a/R/action-link.R
+++ b/R/action-link.R
@@ -16,7 +16,7 @@ add_link_action <- function(trigger, board, update, ...) {
           }
 
           blockr.ui::show_sidebar(
-            "main_sidebar",
+            "actions_sidebar",
             title = "Create new link",
             ui = body
           )
@@ -110,10 +110,10 @@ add_link_action <- function(trigger, board, update, ...) {
               isolate({
                 body <- link_sidebar_body(session$ns, board$board, src_id)
                 if (is.null(body)) {
-                  blockr.ui::hide_sidebar("main_sidebar")
+                  blockr.ui::hide_sidebar("actions_sidebar")
                 } else {
                   blockr.ui::keep_or_hide_sidebar(
-                    "main_sidebar",
+                    "actions_sidebar",
                     title = "Create new link",
                     ui = body
                   )
--- a/R/action-link.R
+++ b/R/action-link.R
@@ -1,11 +1,21 @@
-add_link_action <- function(trigger, board, update, ...) {
+add_link_action <- function(trigger, board, update, ...,
+                            sidebar_id = "main_sidebar") {
 
   new_action(
     function(input, output, session) {
 
       observeEvent(
         trigger(),
-        showModal(link_modal(session$ns, board$board, trigger()))
+        {
+          body <- link_sidebar_body(session$ns, board$board, trigger())
+          if (is.null(body)) return()
+
+          blockr.ui::show_sidebar(
+            sidebar_id,
+            title = "Create new link",
+            ui = body
+          )
+        }
       )
 
       observeEvent(
@@ -82,7 +92,11 @@ add_link_action <- function(trigger, board, update, ...) {
             list(links = list(add = new_lnk))
           )
 
-          removeModal()
+          blockr.ui::keep_or_hide_sidebar(
+            sidebar_id,
+            title = "Create new link",
+            ui = link_sidebar_body(session$ns, board$board, trigger())
+          )
         }
       )
 
--- a/R/action-link.R
+++ b/R/action-link.R
@@ -1,5 +1,4 @@
-add_link_action <- function(trigger, board, update, ...,
-                            sidebar_id = "main_sidebar") {
+add_link_action <- function(trigger, board, update, ...) {
 
   new_action(
     function(input, output, session) {
@@ -11,7 +10,7 @@ add_link_action <- function(trigger, board, update, ...,
           if (is.null(body)) return()
 
           blockr.ui::show_sidebar(
-            sidebar_id,
+            "main_sidebar",
             title = "Create new link",
             ui = body
           )
@@ -93,7 +92,7 @@ add_link_action <- function(trigger, board, update, ...,
           )
 
           blockr.ui::keep_or_hide_sidebar(
-            sidebar_id,
+            "main_sidebar",
             title = "Create new link",
             ui = link_sidebar_body(session$ns, board$board, trigger())
           )
--- a/R/action-sidebar.R
+++ b/R/action-sidebar.R
@@ -1,15 +1,8 @@
-block_modal <- function(ns, board, mode = c("append", "add", "prepend")) {
+block_sidebar_body <- function(ns, board, mode = c("append", "add", "prepend")) {
 
   mode <- match.arg(mode)
   board_block_ids <- board_block_ids(board)
   board_link_ids <- board_link_ids(board)
-
-  title <- switch(
-    mode,
-    append = "Append new block",
-    prepend = "Prepend new block",
-    add = "Add new block"
-  )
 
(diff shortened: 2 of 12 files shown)


R/action-link.R
```diff
@@ -82,7 +92,11 @@ add_link_action <- function(trigger, board, update, ...) {
             list(links = list(add = new_lnk))
           )
 
-          removeModal()
+          blockr.ui::keep_or_hide_sidebar(
+            sidebar_id,
+            title = "Create new link",
+            ui = link_sidebar_body(session$ns, board$board, trigger())
+          )
```
nbenn, line 126: On a successful confirm with the sidebar pinned, this rebuilds the form via `link_sidebar_body(session$ns, board$board, trigger())`. `link_sidebar_body` is both a UI builder and a side-effecting check: if no other block has a free input (`R/action-sidebar.R:98-104`), it calls `notify("No inputs are currently available.", type = "warning")` and returns `NULL`.

The warning is correct on the *initial* open — the user tried to start a link with no valid destination. But in the pinned chain path, this confirm just successfully consumed an input, and there may be none left; the user shouldn't be warned for that. With `ui = NULL`, `keep_or_hide_sidebar` also re-shows the panel with an empty body and the original title.

Suggestion: in the chain path, detect the "nothing left to link" case and close the sidebar silently instead of routing through `link_sidebar_body`. E.g. compute availability inline here and call `hide_sidebar(sidebar_id)` directly when it's zero, otherwise `keep_or_hide_sidebar(...)` as today.

R/board-ui.R
```diff
@@ -1,14 +1,13 @@
...
-
-  opt_ui_or_null <- function(plg, plgs, x) {
-    if (plg %in% names(plgs)) board_ui(id, plgs[[plg]], x)
-  }
-
+board_ui.dock_board <- function(
+  id,
+  x,
+  plugins = board_plugins(x),
+  options = board_options(x),
```
nbenn: The PR removed the only consumer of `options` (the old `options_ui(id, options, ...)` call), but the parameter is still declared here and the call site in `blockr_app_ui.dock_board()` (`R/utils-serve.R`) still forwards a value. That value now goes nowhere.

This is a behavior regression for `serve(board, options = custom_options(...))` — blockr.core's public override hook (`blockr.core/R/utils-serve.R:163-174`). Pre-PR, the override flowed into `options_ui` and rendered. Post-PR, `settings_body()` recomputes `blockr.core::blockr_app_options(x)` independently of what the caller passed, so the override is dropped on the floor.

Fix: thread `options` from `board_ui.dock_board()` into `settings_body()` — e.g. capture it in `settings_observer()` and call `settings_body(id, board$board, options = options)`. Then the recompute in `settings_body`'s default is just the fallback when no caller-supplied value exists.

R/board-ui.R
```diff
@@ -82,17 +80,42 @@ dock_outputs_ui <- function(id, views) {
...
+#'
+#' Returns a tagList containing the same options accordion that the
+#' Bootstrap offcanvas used to render. Called at server time from
+#' `board_server_callback()` when the user clicks the navbar gear, and
+#' passed to `blockr.ui::show_sidebar()`.
+#'
+#' Uses `blockr.core::blockr_app_options(x)` (rather than `board_options(x)`)
+#' so the accordion includes options contributed by blocks on the board and
+#' by registered block constructors — same set `serve()` passes to the
+#' top-level `board_ui()` call.
```
nbenn: This sentence is what flags the bug in #2 (on `board_ui.dock_board()`'s `options` arg): the recomputed `blockr.core::blockr_app_options(x)` equals what `serve()` passes only along the default path. The moment a caller uses `serve(board, options = custom_options(...))`, the override is dropped here.

Good news: `settings_body` already declares `options` as a parameter (line 104). The fix is plumbing only — pass the caller's value into `settings_observer()` and forward to `settings_body(id, board$board, options = options)`. Then this doc is literally true.

R/action-sidebar.R
```diff
@@ -68,29 +64,22 @@ block_modal <- function(ns, board, mode = c("append", "add", "prepend")) {
...
+    visible_fields,
+    toggle_button(
+      ns("block-advanced-options"),
+      ns("block-advanced-toggle")
+    ),
+    advanced_section,
+    confirm_button(
+      inputId = ns(confirm_id),
+      label = button_label
     )
```
nbenn, line 78: The old `*_modal` builders all ended their tagList with `auto_focus_script(ns(selection_id))`, which focused the first selectize once the modal was shown. All three new `*_sidebar_body` builders drop those lines, so on every action open the focus stays on the trigger button — minor but consistent keyboard-UX regression across all six action handlers.

Note that even keeping the call wouldn't have helped: `auto_focus_script` (`R/action-ui.R:308`) hooks `'shown.bs.modal'`, a Bootstrap modal event the sidebar never fires. Fix is one of:

- Replace `auto_focus_script` with a sidebar-aware variant (bound to whatever event blockr.ui's panel fires on open) and re-add to each tagList, or
- Push focus management into the sidebar primitive itself (probably cleaner — every consumer benefits).

Either way, `auto_focus_script` is now unused and should be deleted or updated in the same PR.

R/board-ui.R
```diff
@@ -66,7 +60,11 @@ board_ui.dock_board <- function(id, x, plugins = board_plugins(x),
         id = id,
         board = x
       )
-    )
+    ),
+    # Sidebar mount: fixed top-level DOM id so any deeply-nested action
+    # handler can target it via `blockr.ui::show_sidebar("main_sidebar", ...)`
+    # through `rootScope` walking. Single sidebar per app at v0.
+    blockr.ui::sidebar_ui("main_sidebar", mode = "push")
```
nbenn: Non-blocking — almost certainly lives in blockr.ui, not here.

Reading the R-side primitive in `blockr.ui/R/sidebar.R`, every `show_sidebar()` call sends an `action = "show"` message that replaces the panel's content and title; pin state is managed client-side by the JS bundle. I haven't traced the JS, so I can't be certain — but it looks like pinning the settings panel and then clicking, say, "Add block" (a different caller targeting the same `"main_sidebar"` id) would silently swap the settings form out from under the user. The chromote run cited in the description only covers the *same-caller* chain-while-pinned case, which is preserved.

If that inference holds, my preferred v0 direction would be: at most one pinned panel at any time; a subsequent `show` from a different caller overlays it (transient), and on close of the transient one the pinned panel re-appears. Keeps "pinned" meaning what users expect without tabs or multiple DOM slots. Worth confirming the failure mode in the JS first.

Leaves the door open for richer follow-ups when warranted — e.g. multiple pinned panels surfaced as tabs, or named slots — but that's strictly future work and shouldn't shape the v0 design.

R/board-server.R
```diff
@@ -170,6 +171,33 @@ init_view_docks <- function(views, board, update, triggers, session, dock_mgr) {
...
+  input <- session$input
+  # Strip the trailing "-" so settings_body() can use NS(id, ...) on the
+  # un-namespaced board module id, exactly as `options_ui()` did before.
+  id <- sub("-$", "", session$ns(""))
+
+  observeEvent(
+    input$settings_btn,
+    {
+      blockr.ui::show_sidebar(
+        "main_sidebar",
```
nbenn: Minor: all six action functions (`add_block_action`, `append_block_action`, `prepend_block_action`, `add_link_action`, `add_stack_action`, `edit_stack_action`) accept `sidebar_id = "main_sidebar"` as a parameter. `settings_observer` hard-codes the same string here. Either parameterize this too (and pass through from `board_server_callback`), or drop the parameter from the actions — they all share the singleton DOM id today anyway, so the parameter currently buys nothing real. Consistency either way.

R/board-server.R
```diff
@@ -170,6 +171,33 @@ init_view_docks <- function(views, board, update, triggers, session, dock_mgr) {
...
+#' the shared sidebar via `blockr.ui::show_sidebar()`.
+#'
+#' @param session Shiny session (the board's session).
+#' @param board Reactive board state.
+#' @noRd
+settings_observer <- function(session, board) {
+  input <- session$input
+  # Strip the trailing "-" so settings_body() can use NS(id, ...) on the
+  # un-namespaced board module id, exactly as `options_ui()` did before.
+  id <- sub("-$", "", session$ns(""))
```
nbenn: Minor: relying on Shiny's namespace separator being literally `"-"` is brittle — there's no public API guarantee for it, and a future Shiny change would break this silently. The comment is honest about the hack but doesn't make it less fragile.

A cleaner option: have `settings_body()` accept an `ns` function (e.g. `session$ns`) instead of a bare `id`, and use `ns("board_options")` directly in place of `NS(id, "board_options")`. Then this strip goes away and the function reads more idiomatically. Alternatively, plumb `id` through from `board_server_callback`'s caller so this side doesn't have to reconstruct it at all.

nbenn (commented):
Phase 2 lands the sidebar adoption cleanly along the happy path — the six action handlers map to the new primitive in a mechanical, easy-to-follow way, and the chromote spot-check covers the same-caller chain-while-pinned flow well. One thread that looks like a potential regression, plus a handful of smaller observations.

**Potential regression:**

- `custom_options()` override may no longer reach the rendered panel (#2 on `R/board-ui.R:6`, related #3 on the roxygen). `serve(board, options = custom_options(...))` is a public override path in blockr.core; pre-PR, `options_ui(id, options, ...)` consumed it, post-PR `board_ui.dock_board()`'s `options` parameter is unused and `settings_body()` recomputes `blockr_app_options(x)` from scratch. If I'm reading that right, the fix is plumbing only — `settings_body()` already declares an `options` parameter; threading the caller's value through `settings_observer()` should be enough. Worth verifying.

**Perhaps worth picking up:**

- The pinned-chain in `add_link` looks like it'll re-fire a "No inputs are currently available." warning after a *successful* add when the last available input was just consumed (#1). Branching on availability before chaining would tidy this up.
- Auto-focus on the first selectize seems to go away for every action open (#4). The old `auto_focus_script` is bound to a Bootstrap modal event the sidebar never fires; `auto_focus_script()` itself (`R/action-ui.R:308`) is now unreachable, so probably an update-or-delete-in-place call.
- `settings_observer` hard-codes `"main_sidebar"` while all six action handlers parameterize via `sidebar_id` (#7) — might be nice to pick one direction.
- `id <- sub("-$", "", session$ns(""))` (#8) leans on Shiny's namespace separator being literally `"-"`; an `ns` closure or threaded `id` would feel less brittle.
- The single-sidebar invariant might let a pinned panel get clobbered by an unrelated action click (#6). Almost certainly blockr.ui follow-up, not this PR.

**One thing I'd flag separately**: no new tests. The diff introduces `settings_observer`, six chain-on-pin paths, and the new early-return in `add_link_action`, but `tests/testthat/` is untouched. The 372/372 number only covers code that hasn't changed. Observer-level tests with a stubbed `blockr.ui::show_sidebar` would cover the regression surface — `test-action-block.R`, `test-action-link.R`, `test-action-stack.R`, `test-board-server.R` are natural homes.

Sequencing note: PR is `mergeable_state: blocked` pending blockr.ui#2 anyway, so none of this is on the critical path.

R/action-sidebar.R
DivadNojnarg: I'd keep `auto_focus_script`: used in R/board-server.R:924 in showModal() within `suggest_panels_to_add()` (totally unrelated to this PR).

DivadNojnarg: I'll re-add the selectize focus in blockr.ui instead, however, this will likely drop in future PR when the UI is revamped.

R/board-ui.R
DivadNojnarg: That won't solve the issue but, we'll go for 2 sidebars. 1 for actions and another one for settings. That's cleaner in term of responsibilities.

While checking this one, I noticed that whenever there are 2 sidebars on opposite sides in push mode, the content is only constrained to one side. That was a bug.

R/board-ui.R
DivadNojnarg: @nbenn This issue revealed a serious gap in the blockr.ui implementation. I noticed that `sidebar_ui()` did not accept initial UI content and everything was managed by `show_sidebar()` on the server side, which not was I initially planed. This caused unnecessary UI/server roundtrip, even to render a fixed set of elements like board options. I modified the blockr.ui PR, so we actually have pretty much nothing to change here, for now. For now, because I noticed board options don't serialise or restore correctly on main, so there will be some change to bring in the future, but not for that PR.

PR #125, opened 2026-05-11
Locked mode fixes

### LINKED ISSUE #122: locked state: does not preserve hidden sections
Saving a board with hidden inputs in one block. Restoring in a locked board, input is not hidden on the same block.

### LINKED ISSUE #123: Locked state: show preserve board plugin UI in locked mode
locked mode should allow to save and restore a board. That's currently not the case for the board plugin

### LINKED ISSUE #124: Locked state: panel saved in an unlocked board are still closable when restored in a locked app

DivadNojnarg:
Fixes #123
Fixes #124 
Fixes #122

### DIFF
--- a/R/plugin-block.R
+++ b/R/plugin-block.R
@@ -139,10 +139,6 @@ block_card_title <- function(block, id, info) {
 
 block_card_toggles <- function(blk, ns, ctrl_meta = NULL) {
 
-  if (is_dock_locked()) {
-    return(NULL)
-  }
-
   vals <- c("inputs", "outputs")
   icon_labels <- list(
     as.character(icon("sliders")),
@@ -176,6 +172,15 @@ block_card_toggles <- function(blk, ns, ctrl_meta = NULL) {
     "blockr-section-toggle",
     trimws(gsub("form-group|ms-auto", "", section_toggles$attribs$class))
   )
+
+  # Locked dock: render the widget hidden so `input$collapse_blk_sections`
+  # still seeds the accordion via the observer below and honors any saved
+  # `attr(blk, "visible")` on restore (#122). Hidden (not removed) means
+  # the user can neither see nor toggle it. The tooltip wiring is skipped
+  # because nothing is hoverable.
+  if (is_dock_locked()) {
+    return(div(style = "display: none;", section_toggles))
+  }
 
   tagList(
     section_toggles,
--- a/R/plugin-block.R
+++ b/R/plugin-block.R
@@ -175,9 +175,16 @@ block_card_toggles <- function(blk, ns, ctrl_meta = NULL) {
 
   # Locked dock: render the widget hidden so `input$collapse_blk_sections`
   # still seeds the accordion via the observer below and honors any saved
-  # `attr(blk, "visible")` on restore (#122). Hidden (not removed) means
-  # the user can neither see nor toggle it. The tooltip wiring is skipped
+  # `attr(blk, "visible")` on restore (#122). The tooltip wiring is skipped
   # because nothing is hoverable.
+  #
+  # NOTE: this is a UI-only hide, not a real lock. `display: none` removes the
+  # widget from view but the checkbox group is still a live Shiny input — a
+  # client can flip it via `Shiny.setInputValue()` and the observer below will
+  # happily mutate accordion state. "Lock" here is a UX affordance, not a
+  # server-side trust boundary. The principled fix is to drop the hidden
+  # widget entirely and seed the accordion directly from `attr(blk, "visible")`
+  # on startup, so the observer can be gated on `!is_dock_locked()`. TODO(#TBD).
   if (is_dock_locked()) {
     return(div(style = "display: none;", section_toggles))
   }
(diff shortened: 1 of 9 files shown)


R/utils-dock.R
```diff
@@ -109,9 +109,33 @@ ext_panel_ids <- function(proxy = dock_proxy()) {
...
+  # tabComponent to follow suit (#124).
+  layout <- lock_panels(layout, locked = is_dock_locked())
+
   dockViewR::restore_dock(proxy, unclass(layout))
 }
 
+# Rewrite every panel's `tabComponent` (and clear the now-meaningless
+# `removeCallback`) so closability reflects the current lock state rather
+# than what was persisted at save time.
+lock_panels <- function(layout, locked) {
```
nbenn, line 135: Asymmetric handling of `removeCallback`: when `locked = TRUE` you clear `params$removeCallback`, but when `locked = FALSE` you only flip `tabComponent` back to `"manual"` without reinstating any callback. Combined with #123 (preserve_board now available in locked mode), a new flow is possible: **save while locked → restore while unlocked**. Those panels will have `tabComponent = "manual"` (so close buttons render) but no `removeCallback`, and may close as no-ops depending on dockview's default. Worth either adding a save-locked → restore-unlocked round-trip test, or documenting the reliance on dockview's default. The current `test-layout-class.R` round-trip starts from an unlocked save, which masks this direction.

R/plugin-block.R
```diff
@@ -177,6 +173,15 @@ block_card_toggles <- function(blk, ns, ctrl_meta = NULL) {
     trimws(gsub("form-group|ms-auto", "", section_toggles$attribs$class))
   )
 
+  # Locked dock: render the widget hidden so `input$collapse_blk_sections`
+  # still seeds the accordion via the observer below and honors any saved
+  # `attr(blk, "visible")` on restore (#122). Hidden (not removed) means
+  # the user can neither see nor toggle it. The tooltip wiring is skipped
+  # because nothing is hoverable.
+  if (is_dock_locked()) {
+    return(div(style = "display: none;", section_toggles))
```
nbenn, line 189: The comment above claims "the user can neither see nor toggle it," but `display: none` only hides — the checkbox group is still a live Shiny input and can be flipped via `Shiny.setInputValue`, which fires the existing observer and mutates accordion state in a "locked" session. The comment overstates the guarantee. A principled improvement here would be to gate the observer that consumes `input$collapse_blk_sections` on `is_dock_locked()` so the server refuses the mutation when locked. That fixes this specific case, but only scratches the surface — the same vector applies to every state-mutating observer in the app, so a broader pattern (e.g. structurally not registering mutating observers in locked mode, or a `guarded_observe` helper) deserves its own follow-up.

tests/testthat/test-plugin-block.R
```diff
@@ -95,3 +95,31 @@ test_that("condition ui test", {
...
+
+  blk <- new_dataset_block()
+  attr(blk, "visible") <- "outputs"
+
+  # Unlocked: widget renders visible, with `selected` matching saved attr
+  unlocked <- withr::with_options(
+    list(blockr.dock_is_locked = NULL),
+    block_card_toggles(blk, NS("x"))
+  )
+  expect_false(inherits(unlocked, "shiny.tag"))   # tagList
```
nbenn: Nit: `expect_false(inherits(unlocked, "shiny.tag"))` is checking "is this a tagList, not a tag," which is correct but indirect. `expect_s3_class(unlocked, "shiny.tag.list")` would say the same thing more directly.

nbenn (approved):
LGTM, thanks! A few things worth a second look left inline.

One adjacent observation, not in this PR's diff: the gear button next to the new lock indicator still opens the options offcanvas in locked mode. If any options inside it are write-actions, they should probably be gated alongside the rest — worth a follow-up issue.

DivadNojnarg:
Thanks @nbenn I can look at the 3 above points. Additionally there are other things to consider:

- what to do with extensions in locked mode (DAG, md, ...): currently we can use the DAG in locked mode and add new panel. I believe we should disable them, but it is probably more tricky than it sounds: what does this mean for the layout ... 
- Empty views are also editable (we can add panel). 

I propose to handle this is another PR once we discussed.

R/utils-dock.R
DivadNojnarg: @nbenn I propose to make things symmetrical. We adjust the callback setup according to the dock mode: if locked, custom tab component and no callback. If unlocked then, tab component manual and callback default to the one provided by dockViewR (which is fine for now, we don't need to provide our own callback).

```r
lock_panels <- function(layout, locked) {
  if (isTRUE(locked)) {
    tab_component <- "custom"
    remove_callback <- NULL
  } else {
    tab_component <- "manual"
    remove_callback <- list(
      `__IS_FUNCTION__` = TRUE,
      source = unclass(dockViewR::default_remove_tab_callback())
    )
  }
  layout[["panels"]] <- lapply(layout[["panels"]], function(p) {
    p[["tabComponent"]] <- tab_component
    p[["params"]][["removeCallback"]] <- remove_callback
    p
  })
  layout
}
```

Then we add the missing test to cover this case. How does this sound to you?

R/plugin-block.R
DivadNojnarg: @nbenn For now, I can update the comment to make it fair about the intentions. The real fix may involve gating other elements, so to find a generalisable solution, probably out of scope of this PR if we want something nice.

Issue #126, opened 2026-05-11
Locked state: what to do with extensions?


DivadNojnarg:
We can still interact with DAG in locked mode which basically allows to add new panels to the layout. This should be disabled.
Beyond the DAG, this is more an extension issue. What do we do with extensions in locked mode? I believe we should not show them.

PR #137, opened 2026-05-26
Fix dock locked part2

### LINKED ISSUE #127: Locked state: gate observers
Arises in https://github.com/BristolMyersSquibb/blockr.dock/pull/125#discussion_r3260364038

It's not enough to just hide UI elements in locked mode. We also have to act on the server side to lock the observers.

### LINKED ISSUE #135: Locked state: manage options access
Follow up of #125

### LINKED ISSUE #136: Locked state: prevent new panel addition in empty views
Follow up of #125

DivadNojnarg:
fix #127 
fix #136 
fix #135

#126 is left to extensions's maintainers: blockr.dock exports helpers such as `is_dock_locked()`, but how to use them, only extensions know.


DivadNojnarg:
Also the runtime `is_dock_locked()` (being able to lock/unlock at runtime) is deferred to another PR. This needs more thoughts.

nbenn:
Superseded by #249, which landed the agreed approach and auto-closed #127, #135 and #136.

The design settled on putting the trust boundary in blockr.core — a single deny-all `board_update` gate plus a per-block `frozen` channel — and having blockr.dock consume it, rather than gating each mutating observer individually with `req_unlocked()`. Every action handler already routes through `update()`, so core's gate now rejects those mutations on a locked board regardless of emitter, which makes the per-observer gates here redundant. `is_dock_locked()` also stays internal — extensions read core's exported `is_board_locked()` — so #126 needs no dock-side helper.

The one idea worth carrying over from the efficiency discussion — gating *creation* of the observers that still fire futilely on a locked board (chiefly the grid-echo layout mirror) — is tracked separately in #339.

Closing as superseded. Thanks for the groundwork here — it shaped the final design.

PR #147, opened 2026-05-26
Expose view_data to extensions; add as_layout_spec()

### LINKED ISSUE #145: Add as_layout_spec() to reconstruct a layout's constructor form
A stored `dock_layout` is represented internally as a tree of `{type, data, size}` nodes (the dockview wire format). To inspect or modify a complex nested layout, callers need to see it in the same vocabulary they use at construction time — `dock_layout(...)` with `panels()` and `group()` children. Today there is no inverse for the lowering that the constructors do; you can write a layout but you can't easily read one back in its written form.

Motivation surfaced in BristolMyersSquibb/blockr.dock#141: for an external mutator (LLM, programmatic caller) to make a meaningful modification to a complex view, it has to be able to *see* the current view in a form it can reason about and edit. The alternative — exposing raw tree nodes — leaks the dockview wire format into every consumer.

## Proposal

A pure tree-walker `as_layout_spec(layout)` that converts a stored `dock_layout` back into spec form. Idempotent: `as_layout_spec(dock_layout(...))` returns a `dock_layout` equivalent to the input.

```r
as_layout_spec <- function(layout) {
  stopifnot(is_dock_layout(layout))

  walk <- function(node) {

    if (identical(node$type, "leaf")) {

      views <- unlist(node$data$views, use.names = FALSE)
      active <- node$data$activeView

      if (length(views) == 1L) {
        return(views)
      }
      if (identical(active, views[[1L]])) {
        return(views)            # bare character vector
      }
      return(panels(!!!views, active = active))
    }

    # branch
    children <- lapply(node$data, walk)
    sizes <- vapply(node$data, function(x) x$size %||% NA_real_, numeric(1L))

    if (all(is.na(sizes)) || length(sizes) <= 1L ||
          all(abs(sizes - mean(sizes)) < 1e-9)) {
      return(children)            # bare list, even split
    }

    group(!!!children, sizes = sizes)
  }

  root <- layout$grid$root
  if (is.null(root)) return(dock_layout())

  children <- if (identical(root$type, "branch")) lapply(root$data, walk) else list(walk(root))

  do.call(
    dock_layout,
    c(
      children,
      list(
        orientation  = tolower(layout$grid$orientation),
        active_group = layout$activeGroup
      )
    )
  )
}
```

(Sketch — actual implementation will need to handle empty branches, edge cases around `sizes` ≈ even, and propagate `active` attribute if the layout is marked as the active view.)

## Edge cases to verify

- **Empty layout**: `dock_layout()` round-trips to itself.
- **Single-panel leaf**: round-trips to a bare string.
- **Multi-panel leaf with default active**: round-trips to a character vector (not a `panels()` call).
- **Multi-panel leaf with explicit active**: round-trips to `panels(..., active = ...)`.
- **Branch with even sizes**: round-trips to a bare list (not `group(..., sizes = c(...))`).
- **Branch with uneven sizes**: round-trips to `group(..., sizes = c(...))`.
- **Nested branches**: orientations alternate automatically; only the top-level orientation needs to be specified explicitly.
- **`activeGroup`**: carried through.

## Optional: `print.dock_layout` method

Use `as_layout_spec()` to produce a readable printout:

```r
print.dock_layout <- function(x, ...) {
  spec <- as_layout_spec(x)
  cat(deparse(spec_to_call(spec)), sep = "\\n")   # or similar
  invisible(x)
}
```

Worth adding once `as_layout_spec` lands and is solid. Can be a follow-up.

## Tests

A `test-layout-class.R` round-trip table:

```r
expect_equal(as_layout_spec(dock_layout("a")),       dock_layout("a"))
expect_equal(as_layout_spec(dock_layout("a", "b")), dock_layout("a", "b"))
expect_equal(
  as_layout_spec(dock_layout(panels("a", "b", active = "b"))),
  dock_layout(panels("a", "b", active = "b"))
)
expect_equal(
  as_layout_spec(dock_layout("a", "b", sizes = c(0.3, 0.7))),
  dock_layout("a", "b", sizes = c(0.3, 0.7))
)
expect_equal(
  as_layout_spec(
    dock_layout("a", group("b", "c", sizes = c(0.6, 0.4)), sizes = c(0.3, 0.7))
  ),
  dock_layout("a", group("b", "c", sizes = c(0.6, 0.4)), sizes = c(0.3, 0.7))
)
```

`identical()` may need to allow for minor structural differences (e.g. `as.list` vs `c()` on character vectors); equality should be on the validated lowered form.

## Non-goals

- No new payload-level capability — this is a pure read-side ergonomics fix.
- No introduction of stable IDs for branches. Dockview's wire format doesn't carry them and adding parallel ID infrastructure would require fighting the underlying library. See BristolMyersSquibb/blockr.dock#141 for the discussion.
- No mutation surface on the spec form itself. Callers manipulate the result with normal R list/spec operations and pass it back via `dock_layout(...)` or directly as a payload value.

## Downstream

- BristolMyersSquibb/blockr.dock#141's `list_views`-style read tool returns `as_layout_spec(...)` outputs so callers see constructor-shaped values.
- Useful for documentation: examples can print real layouts inline.
- Useful for debugging: `print.dock_layout` becomes meaningful.

Builds on the data model landed by #115. No dependency on the augment / apply lifecycle work.

### LINKED ISSUE #146: Expose live layouts to extensions via view_data
Extension servers today receive an active-view-only handle (`dock = dock_mgr$active_dock`) but no read access to the full live `dock_layouts` collection. The reactive that holds this — `live_view_data` — already exists in `R/board-server.R:230`; it's just not threaded through to extensions.

This blocks any extension that needs to reason about the live layout of any view other than the active one (e.g. "which view currently contains panel X?", "list all views and their panel arrangements"). The immediate downstream consumer is the LLM tool surface in blockr.assistant (see BristolMyersSquibb/blockr.dock#141), but the change is independent and useful on its own.

## Current state

In `board_server_callback` (`R/board-server.R:17`):

```r
view_data <- live_view_data(vs, dock_mgr)

ext_res <- lapply(
  exts,
  extension_server,
  list(board = board, update = update, dock = dock, actions = triggers),
  list(...)
)
```

`view_data` is returned at the top level of `board_server_callback` but is not in the list passed to extension servers.

## Proposal

Add `view_data` to the extension server's argument list:

```r
ext_res <- lapply(
  exts,
  extension_server,
  list(
    board     = board,
    update    = update,
    dock      = dock,
    actions   = triggers,
    view_data = view_data
  ),
  list(...)
)
```

Extensions that don't use it ignore the new argument (the per-extension server function receives it via `...` and can take or ignore it). Existing extensions continue to work unchanged.

## Semantics

- `view_data` is a `shiny::reactive()` returning a `dock_layouts` object — names, active marker, and per-view live grids.
- Invalidation: fires when any view's live layout changes (drag, resize, panel add/remove via UI) or when view membership changes (CRUD).
- Read pattern in an extension:

```r
my_extension_srv <- function(id, board, update, ..., view_data) {
  moduleServer(id, function(input, output, session) {
    observe({
      layouts <- view_data()
      # layouts is a dock_layouts; iterate views, inspect grids
    })
  })
}
```

## Documentation

- `R/ext-class.R`'s `extension_server` Rd should document the new argument name and shape.
- A short note in the package vignette covering the extension contract.

## Tests

- Update existing extension fixtures used in `test-board-server.R` (if any) to demonstrate `view_data` access.
- Add a fixture extension that asserts `is_dock_layouts(view_data())` and that `names(view_data())` matches `names(board_layouts(board$board))` after view CRUD events.

## Non-goals

- No new mutation surface for extensions. Writes still go through `update()`.
- No subsetting / per-view filtering — extensions receive the full collection and pick what they need.
- No deprecation of `dock = dock_mgr$active_dock`. That handle remains useful for extensions that only care about the active view.

## Downstream

- Unblocks BristolMyersSquibb/blockr.dock#141's read-side tool surface (`list_views`-style operations consume `view_data()`).
- Future extensions (DAG views, layout previews, sidebar mini-maps) get the same affordance for free.

nbenn:
## Summary

Two layout-facing changes, formerly stacked as #148 → #147; #148 was closed and its commit folded here after the stack landed on main.

- **`view_data` for extensions** (`R/board-server.R`): thread the existing `live_view_data()` reactive into each extension server's named arg list, so extensions can read the full `dock_layouts` of every live view, not just the active one. Extensions absorbing extra args via `...` are unaffected.
- **`as_layout_spec()`** (`R/layout-class.R`): reconstructs a stored `dock_layout` into canonical constructor form. Rewritten against main's post-#149 conversion API — it's now `set_active_view(as_dock_layout(as.list(layout)), is_active_view(layout))` rather than a hand-rolled grid walker, since `as.list()`/`as_dock_layout()` already normalise sizes to ratios and drop even-splits.

Fixes #146
Fixes #145


nbenn:
Closing without merging. `as_layout_spec()` has no real consumer:

- blockr.dock#151's `layouts_match()` can compare via `as.list()` (the canonical spec form — sizes already normalised to ratios, even-splits dropped) instead; the round-trip back to a `dock_layout` adds nothing.
- blockr.assistant#19 only references it in a doc comment and ships its own spec walker; it does not call it.

The `view_data` plumbing this PR also carried is needed (blockr.assistant#19) and is already inherited by #151 (based on this branch), so it folds forward there — same pattern as #148's `as_layout_spec` folding into #147. Branch retained as #151's base.

Issue #141, opened 2026-05-26
External control for the dock layout manager


nbenn:
Companion to blockr.core's per-block external-control pattern (`ctrl_block` plugin + `new_block(external_ctrl = …)`): expose a documented imperative surface for driving the dock layout manager from outside the dockview UI, so extensions like blockr.assistant can place panels, switch views, and manage views programmatically.

Builds on the data-model cleanup in #115. That PR turns `dock_layout` into a small serializable tree (`{grid, activeGroup}`) and lands the declarative setters `active_layout(brd) <-`, `board_layouts(brd) <-`, `active_view(brd) <-`. The gap that remains is the **live, in-session** side: those board-level setters mutate the board object but have no effect on the running dockview proxy mid-session, and the user-driven UI flows for view CRUD have no programmatic counterpart.

## What's missing today

- **View CRUD is UI-only.** `add_view_observer`, `remove_view_observer`, `rename_view_observer` in `R/board-server.R` are wired to `input$view_nav_*`. No `add_view()` / `remove_view()` / `rename_view()` / `set_active_view()` callable from R.
- **Panel moves are dockViewR-only.** `dockViewR::move_panel(id, position, group, index)` and `move_group(from, to, position)` exist underneath but have no `blockr.dock` wrappers (compare: `add_block_panel`, `remove_block_panel`, `select_block_panel` already wrap their dockViewR counterparts).
- **Extensions see one view's dock.** The extension server receives `dock = dock_mgr\$active_dock` — a `reactiveValues` mirror of the *active view's* dock module result. Multi-view access (other views' proxies, view-switch hooks) is not part of the public extension contract.

### 1. Layout controller handle for extensions

A stable handle passed to extension servers alongside today's `dock` (which we'd keep, as the active-view convenience), exposing:

- `views()` — view names + which is active
- `view_layout(name)` — reactive returning that view's live `dock_layout` (mirrors today's `live_view_data` but per-view)
- `view_proxy(name)` — the view's dockViewR proxy, for callers that want raw access
- `set_active_view(name)` — programmatic equivalent of clicking a view tab
- `add_view(name, layout)` / `remove_view(name)` / `rename_view(from, to)` — programmatic counterparts of the UI observers

Working name: \`dock_controller\` (bikeshed welcome).

### 2. Live panel-mutation wrappers

Companions to the existing `add_block_panel` / `remove_block_panel` / `select_block_panel` family, wrapping the remaining dockViewR proxy methods:

[...]

PR #148, opened 2026-05-26
Add as_layout_spec() to reconstruct a layout's constructor form

### LINKED ISSUE #145: Add as_layout_spec() to reconstruct a layout's constructor form

nbenn:
## Summary

- Add `as_layout_spec(layout)`: canonicalises a stored `dock_layout` by routing it through the spec form (`as.list()` / `layout_to_spec()`) and back, preserving the active-view marker.
- Sizes normalise to ratios, even splits and default-active tabs collapse to bare form, and dockview's volatile leaf IDs are regenerated — so `as_layout_spec(dock_layout(...))` round-trips.
- Reuses the spec serdes layer that landed via #149 rather than re-walking the dockview grid, so there's no duplicated normalisation logic.
- Enables programmatic and LLM-driven consumers (see #141) to read a complex view in the same vocabulary they use to write one, without leaking the dockview wire tree.

Fixes #145


nbenn:
Closing without merging — `as_layout_spec()` is redundant with what #149 landed on `main`.

While this PR iterated, #149 introduced the spec round-trip: `as.list.dock_layout()` (→ `layout_to_spec()`) and `as_dock_layout.list()` (→ `spec_to_layout()`), plus `layout_to_json()`/`layout_from_json()`. That already delivers the read-a-layout-in-constructor-vocabulary ergonomics #145 asked for. `as_layout_spec(x)` had collapsed to `as_dock_layout(as.list(x))` (plus active-view preservation), so exporting it would just add a confusing near-synonym.

Issue #145 is closed as delivered by #149. The one downstream consumer, `layouts_match()` in #151, has a migration note: #151 (comment), swap `as_layout_spec(ly)` → `as.list(ly)`. #147 only inherited the definition (no call sites) and needs nothing beyond a retarget to `main`.

Leaving the `145-as-layout-spec` branch in place for now since #147 is still based on it.

PR #151, opened 2026-05-27
Implement structured views slot in board_update payload

### LINKED ISSUE #150: Implement views slot in board_update payload via extension generics
Implement the `views` slot in the `board_update` payload via the extension generics. This is the substantive delta work — it's what makes layouts mutable through the same channel as blocks / links / stacks, with atomic cross-slot transactions.

Design and rationale: BristolMyersSquibb/blockr.dock#141. Foundation: BristolMyersSquibb/blockr.core#184.

## Payload contract

After this lands, payloads can include a `views` slot alongside the existing `blocks` / `links` / `stacks` slots:

```r
list(
  blocks = list(add = ..., mod = ..., rm = ...),    # unchanged
  links  = list(add = ..., mod = ..., rm = ...),    # unchanged
  stacks = list(add = ..., mod = ..., rm = ...),    # unchanged
  views  = list(
    add    = list(NewView = dock_layout(...), ...), # new views, full layout
    mod    = list(Existing = dock_layout(...), ...),# full desired layout per view
    rm     = c("OldView"),                          # views to remove
    active = "SomeView"                             # optional active-view switch
  )
)
```

Semantics:

- `views$mod[[v]]` is a **full `dock_layout`** — the desired new state for that view. The receiver computes the diff against the current live layout and picks an application strategy.
- Only mentioned views are touched. Omitted views keep their current state.
- Within a payload: `add` and `rm` are disjoint by view name; `mod` cannot reference a name in `rm` or `add`; `active` must resolve to (current ∪ `add`) − `rm`; panel-ID references in `add` and `mod` must resolve to (current blocks ∪ `blocks$add`) − `blocks$rm`.

### 1. `augment_board_update.dock_board`

S3 method on `dock_board`. Calls `NextMethod()` to run core's augmentation first, then:

- **Validates the `views` slice** (structural — well-formed names, no in-payload conflicts; xref — panel IDs resolve in the augmented post-state). Throws classed errors on malformed input.
- **Auto-augments view layouts when blocks are removed.** For every `blocks$rm = ids`, walk current view layouts (from `board_layouts(board)` — assumes #144 has landed and the spec is current). For any view containing affected panel IDs, generate or extend a `views$mod[[v]]` entry that drops those IDs from the layout, preserving the rest of the structure.

```r
augment_board_update.dock_board <- function(upd, board) {

  upd <- NextMethod()

  if (length(upd$blocks$rm)) {
    upd$views$mod <- merge_views_mod(
      upd$views$mod,
      drop_panels_from_layouts(board_layouts(board), upd$blocks$rm)
    )
  }

  if (length(upd$views)) {
    validate_views_delta(upd$views, board, upd)   # post-augment state for xrefs
  }

  upd
}
```

### 2. `apply_board_update.dock_board`

S3 method on `dock_board`. Calls `NextMethod()` to run core's apply first (so newly-added blocks / extensions exist before view mods reference them), then applies the `views` slice in this order:

1. `views$rm` — destroy per-view dock modules, remove DOM containers, drop from `vs$state` and `board_layouts(rv$board)`.
2. `views$add` — insert UI containers, start `manage_dock` modules, update `vs$state` and `board_layouts(rv$board)`.
3. `views$mod` — for each view, compute diff against current live layout and apply (v1: always `restore_layout`; surgical paths deferred — see §4 below).
4. `views$active` — emit the `switch-view` custom message, update `vs$state` active marker, swap `dock_mgr$active_dock`.

```r
apply_board_update.dock_board <- function(board, upd, rv, ...) {

  NextMethod()

  if (length(upd$views$rm))    apply_views_rm(upd$views$rm, rv, ...)
  if (length(upd$views$add))   apply_views_add(upd$views$add, rv, ...)
  if (length(upd$views$mod))   apply_views_mod(upd$views$mod, rv, ...)
  if (!is.null(upd$views$active)) apply_views_active(upd$views$active, rv, ...)
}
```

Each `apply_views_*` helper takes the slice + `rv` (to update `board_layouts(rv$board)`) + whatever dock-manager handle it needs to reach the per-view proxies.

### 3. Internal helpers

- `validate_views_delta(views, board, payload)` — structural check + xref check against the augmented payload's post-state.
- `drop_panels_from_layouts(layouts, ids)` — returns a named list of view-name → modified `dock_layout` for every view whose grid currently contains any of `ids`. Drops affected panel IDs; empties leaves / branches recursively. Returns empty for unaffected views.
- `merge_views_mod(user_mod, augment_mod)` — when both target the same view, augment runs on top of the user's submitted layout (drop IDs from whatever the user supplied). When only one is present, that one wins.
- `apply_views_rm`, `apply_views_add`, `apply_views_mod`, `apply_views_active` — see §2.
- A diff entry point `apply_layout_diff(view, current, target, ctx)` that v1 implements as unconditional `restore_layout`. v2 can dispatch on diff shape.

### 4. Diff machinery — v1 vs later

v1: `apply_views_mod` calls `restore_layout(target, proxy, blocks, extensions)` unconditionally. Correct, slightly visually jarring on every change.

Deferred (separate follow-up issue once v1 is in): surgical paths behind `apply_layout_diff`. Cheap wins:

- Identical layout → noop.
- Only `activeGroup` differs → set active group via proxy, no rebuild.
- Identical structure with one leaf's `activeView` changed → `dockViewR::select_panel(new_active, proxy)`.
- Structurally identical but reordered panels within / across leaves → `dockViewR::move_panel(id, position, group)` per delta.
- Anything else → `restore_layout` fallback.

The diff is computed once per `mod[[v]]` at apply time; the surface above this stays unchanged.

## Source-of-truth assumption

This issue assumes #144 has landed (or lands alongside). Specifically:

- `board_layouts(rv$board)` is the source of truth for current view membership and per-view grids.
- `augment_board_update.dock_board` reads from `board_layouts(board)` (where `board` is the un-mutated `rv$board` at the +Inf moment) when computing view cleanup for `blocks$rm`.
- `apply_board_update.dock_board` writes back to `board_layouts(rv$board)` after each mutation, so the next +Inf observer sees current state.

If #144 hasn't landed yet, fall back to reading from `live_view_data()` snapshots — but the spec-vs-live divergence makes that fragile. Better to gate this issue on #144.

## Structural rejections

`validate_views_delta` must reject:

- `add` and `rm` containing the same view name.
- `mod` referencing a name in `rm`.
- `mod` referencing a name in `add` (caller should fold into `add`).
- `active` referencing a name in `rm` and not in `add`.
- `add[[v]]` or `mod[[v]]` referencing panel IDs that don't resolve in (current blocks ∪ `blocks$add`) − `blocks$rm`.
- Any layout value that fails `validate_dock_layout(x, blocks)`.

Errors use classed conditions (e.g. `dock_views_delta_add_rm_clash`, `dock_views_delta_panel_ref_invalid`).

## Tests

- **Round-trip**: starting from a known `dock_layouts`, sending a payload with each of `add` / `mod` / `rm` / `active` arrives at the expected `board_layouts(rv$board)`.
- **Atomicity with blocks**: a payload with `blocks$add = new1` + `views$mod[[v]] = dock_layout("new1")` applies in the right order; `new1` exists by the time the view mod runs.
- **Block-removal cleanup**: a payload with only `blocks$rm = "b"` results in every view containing `b` having `b` dropped from its layout (no orphan refs reach dockview).
- **Augment + user mod merge**: payload with both `blocks$rm = "b"` and `views$mod[["Analysis"]] = <user layout containing "b">` — `b` is dropped from the merged result.
- **Structural rejections**: every classed error path covered.
- **Reset**: after a `views`-only payload applies, `board_update()` is `NULL`.

## Cleanup: drop the `rm_blocks.dock_board` workaround

Once `augment_board_update.dock_board` is doing surgical block-removal cleanup, the temporary workaround in `R/dock-board.R:210` is dead code and should be removed in this same change:

```r
# Current — to be deleted:
rm_blocks.dock_board <- function(x, rm, ...) {
  # for now, b/c dock$layout(), passed in ... is not updated in time, we can
  # only clear the layout to not run into validation issues
  active_layout(x) <- resolve_dock_layout(extensions = dock_extensions(x))
  NextMethod(object = x)
}
```

The augment hook handles every view (not just the active one) and preserves the rest of each affected view's layout (not just nuking it). Strictly better behavior; bundled here because the workaround only exists *because* of the gap this issue closes.

**NEWS entry** calls out the visible behavior change: "Removing a block no longer clears the active view's layout. Instead, every view containing the removed block has the block's panel dropped surgically, preserving the remaining structure."

## Non-goals

- Surgical diff paths inside `apply_views_mod`. v1 = `restore_layout` always. Surgical paths are a follow-up.
- LLM / blockr.assistant tool surface. Downstream consumer, separate package, separate issue.
- Any change to view-CRUD UI flows. Programmatic CRUD is added; the existing UI observers stay as-is (and could later be refactored to write into the same payload format, but that's out of scope here).

## Dependencies

- **Blocks on**: BristolMyersSquibb/blockr.core#184 (extension generics).
- **Strongly prefers**: #144 (sync `board_layouts` with live state) lands first or alongside.
- **Unlocks**: blockr.assistant tool surface.

nbenn:
## Summary

- `views` slot in `board_update` becomes a structured delta (`add` / `mod` / `rm` / `active`) instead of a wholesale `dock_layouts` replacement. Mentioned views are touched, omitted views keep their current state, and the four sub-slots compose atomically with `blocks` / `links` / `stacks` in the same lifecycle tick.
- `augment_board_update.dock_board()` validates the delta and auto-augments `views$mod` for every view whose layout contains a panel in `blocks$rm`, dropping the affected panel IDs while preserving the rest of the grid. `merge_views_mod()` composes the augment on top of any user-submitted `views$mod` for the same view.
- `apply_board_update.dock_board()` applies the slice in the spec order (`rm`, `add`, `mod`, `active`) after `NextMethod()` runs core's apply, so newly-added blocks exist before view layouts reference them. `apply_views_mod` short-circuits `restore_layout` when the live dockView state already matches the target — required to break the sync→apply→restore feedback loop the UI mirror creates on every drag-end.
- `board_server_callback` returns `dock_mgr` (carrying `update`, `triggers`, `board_rv`, `vs` references) so the `-Inf` apply observer can reach per-view dock proxies for `add` / `rm` / `mod` / `active` UI mutations.
- `rm_blocks.dock_board` now drops affected panels surgically from every view (instead of nuking the active view's layout entirely) for direct callers like `clear_board()`. The augment hook handles the lifecycle path.

Stacked on #147. Retargets to `main` once that lands.

Fixes #150


nbenn:
## Heads-up: `as_layout_spec()` is being dropped — migrate `layouts_match()` to `as.list()`

Context for whoever picks up this branch. PR #148 (`145-as-layout-spec`, the base further down this stack) is **not** going to add `as_layout_spec()` after all. While #148 was iterating, #149 landed the spec round-trip on `main` — `as.list.dock_layout()` (→ `layout_to_spec()`) and `as_dock_layout.list()` (→ `spec_to_layout()`), plus `layout_to_json()`/`layout_from_json()`. That already delivers the read-in-constructor-vocabulary ergonomics #145 asked for, so a separate `as_layout_spec()` is just a near-synonym for `as_dock_layout(as.list(x))` and we're closing #145/#148 rather than exporting it.

**Impact on this PR:** the only consumer is `layouts_match()` in `R/views-delta.R` (added by e95a94e). When this stack rebases onto `main`, the inherited `as_layout_spec()` definition (in `R/layout-class.R`, its `NAMESPACE` export, and the `test-layout-class.R` round-trip tests) disappears with it — all inherited from #148's branch, none of it added here, so just let it drop on rebase.

**Migration** — `as.list()` is actually a better fit for the canonical-equality check than `as_layout_spec()` was:

```r
layouts_match <- function(a, b) {

  to_spec <- function(x) {

    if (is.null(x)) {
      return(NULL)
    }

    ly <- tryCatch(as_dock_layout(x), error = function(e) NULL)

    if (is.null(ly)) {
      return(NULL)
    }

    spec <- tryCatch(as.list(ly), error = function(e) NULL)

    if (is.null(spec)) {
      return(NULL)
    }

    spec[["focus"]] <- NULL   # ignore focused-panel pointer
    spec
  }

  spec_a <- to_spec(a)
  spec_b <- to_spec(b)

  if (is.null(spec_a) || is.null(spec_b)) {
    return(FALSE)
  }

  identical(spec_a, spec_b)
}
```

What changes vs the current helper:

- `as_layout_spec(ly)` → `as.list(ly)`. Returns the canonical spec list directly (sizes normalised to ratios, even splits and default-active tabs collapsed).
- Drop `attr(spec, "active") <- NULL` — the spec list has no `active` attribute (that's a `dock_layouts`-collection concern, never on a single layout's spec).
- Drop `spec[["activeGroup"]] <- NULL`, **strip `spec[["focus"]]` instead.** The spec omits dockview's volatile per-leaf group IDs entirely, so there's no `activeGroup` to ignore; the only focus-equivalent field is `focus` (the focused-panel pointer `layout_to_spec()` derives from `activeGroup`).

Verified against `pkgload::load_all()` on current `main` + this change:

- two layouts differing only in focused group → match `TRUE`
- pixel-count sizes vs the equivalent ratios (e.g. 300/700 vs 0.3/0.7) → match `TRUE` (canonicalisation normalises both)
- genuinely different structure → match `FALSE`

PR #155, opened 2026-05-28
feat: adopt the block browser in add / append / prepend actions

DivadNojnarg:
## Summary

Adopts the new `blockr.ui` block browser (PR2 of the `add-block-browser` openspec change). The three action handlers (`add_block_action`, `append_block_action`, `prepend_block_action`) now mount the block-browser Shiny module, pass an appropriate target descriptor, and observe its committed-block reactive instead of the per-field input stream.

- **Action handlers (`R/action-block.R`):** mount `blockr.ui::block_browser_server(\"browser\")` once per handler; on trigger, `show_sidebar(ui = blockr.ui::block_browser_ui(session\$ns(\"browser\"), board\$board, target))` where `target` is `NULL` (add), `append_to(trigger())`, or `prepend_to(trigger())`. A single `observeEvent(added(), ...)` builds the block + link(s) and applies one `update()`.
- **`block_sidebar_body()` removed:** no in-tree callers remain. Link / stack flows continue via `link_sidebar_body()` / `stack_sidebar_body()` unchanged.
- **Append fallback:** when the browser hides the input-port picker (single named slot, or variadic block such as `rbind_block`), the handler falls back to `block_input_select(new_blk, ..., mode = \"inputs\")[1L]` so the link wires correctly without user input.
- **Per-field inputs removed:** the old `<mode>_block_selection` / `<mode>_block_id` / `<mode>_block_name` / `<mode>_link_id` / `<mode>_block_input` / `<mode>_block_confirm` are gone in favour of `blockr.ui::block_browser_server()`'s returned reactive.

### Dependency

`DESCRIPTION` pins blockr.ui's Remotes at `BristolMyersSquibb/blockr.ui@add-block-browser` while the matching upstream PR (#6) is open. Once that PR merges, the ref can be dropped.

## Test plan

- [ ] \`devtools::test()\` - rewritten \`test-action-block.R\` drives \`session\$setInputs(\`browser-commit\` = list(...))\` with structured specs (mocked blockr.ui sidebar fns) and covers add, append, prepend, the new fallback for single-input new blocks, and the chain via \`keep_or_hide_sidebar()\` on confirm
- [ ] \`lintr::lint_package()\` clean
- [ ] \`devtools::check()\` 0 errors / 0 warnings
- [ ] Manual: add / append / prepend flows in the dock + dag example


nbenn (approved):
Lgtm, thanks!

PR #23, opened 2025-10-27
Add dock extension panel

nbenn:
Fixes #21.

PR #163, opened 2026-05-31
Add external control for dock extensions

### LINKED ISSUE #162: External control for dock extensions (reactiveVal state + extensions slot in board_update)
## Motivation

`blockr.assistant` can CRUD blocks, links, stacks, and views by staging `mod` deltas on the `board_update` payload. Extensions are the one board-level concept it can't touch: an extension's `state` is exposed read-only, so e.g. the assistant cannot edit a `blockr.md` document's text. There is no write path into extension state at all.

This ports blockr.core's **block external control** to extensions, unchanged in shape:

- Opt-in: `external_ctrl = FALSE | TRUE | c("arg", ...)` on `new_block()` (blockr.core `R/block-class.R:164`), read back by `block_external_ctrl_vars()` (`:670`).
- Controllable `state` entries are `reactiveVal`s — `block_server.block` hard-requires it (`R/block-server.R:168-179`).
- A `blocks$mod` delta lands via `apply_block_mod_delta()`, which writes the live `reactiveVal` (`R/board-server.R:406-429`); `validate_board_update()` gates the keys against `block_external_ctrl_vars()`.

The structure is already 1:1 for extensions — the round-trip proves it. `serialize_board.dock_board` (`R/utils-serdes.R:2`) reads each extension's live `$state` and threads it into `blockr_ser`, and `blockr_deser.dock_extension` (`:111`) rebuilds via `ctor(payload)`. That only works because an extension's `state` names already correspond to its constructor arguments — exactly like blocks. External control is just the **write** face of the same `state` the serializer already **reads**.

This is **opt-in and non-breaking**: extensions default to `external_ctrl = FALSE` and are unaffected.

### 1. Opt-in declaration

`new_dock_extension()` gains `external_ctrl = FALSE` (`TRUE` | character vector), stored as an attribute. Add `extension_external_ctrl_vars(x)`, mirroring `block_external_ctrl_vars()` — resolving `TRUE` **statically** against the extension's constructor inputs (`extension_ctor(x)` formals), since `state` ↔ ctor-args already holds:

```r
extension_external_ctrl_vars <- function(x) {

  res <- attr(x, "external_ctrl")

  if (isTRUE(res)) {
    ext_ctor_inputs(x)
  } else if (isFALSE(res)) {
    character()
  } else {
    stopifnot(is.character(res), all(res %in% ext_ctor_inputs(x)))
    res
  }
}
```

`ext_ctor_inputs()` is the extension analog of blockr.core's `block_ctor_inputs()` (`setdiff(names(formals(<ctor>)), "...")`), derived from `extension_ctor(x)`.

### 2. State promotion + enforcement

Controllable entries of the returned `state` list must be `reactiveVal`s (read-write), not plain `reactive`s. Enforce in `validate_ext_srv_result()` (`R/ext-class.R:164`), which already validates the server return:

```r
vars <- extension_external_ctrl_vars(x)

miss <- setdiff(vars, names(res$state))
if (length(miss)) {
  blockr_abort(
    "Externally controllable variable{?s} {miss} absent from extension `state`.",
    class = "extension_ctrl_var_absent"
  )
}

if (!all(lgl_ply(res$state[vars], inherits, "reactiveVal"))) {
  blockr_abort(
    "Externally controllable extension state must be `reactiveVal`.",
    class = "extension_ctrl_var_not_rv"
  )
}
```

The promotion is backward-compatible with serialization: `reval_if` reads a `reactiveVal` exactly as it reads a `reactive`, so the round-trip is unaffected.

### 3. Author contract

The extension's server must observe its controllable `reactiveVal`s and reflect them into its UI — the same `state -> updateXxx` sync the subset block does (blockr.core `R/transform-subset.R`). With no expr/eval pipeline, this observer is the entire apply mechanism on the extension side. (This is the same responsibility a block author already carries to avoid client/server state drift — not new.)

### 4. Transport — `extensions$mod` slot

```r
list(extensions = list(mod = list(document = list(content = "# rewritten"))))
```

- `validate_board_update.dock_board` (`R/dock-board.R:294`) — structural check; gate each `mod[[id]]` key against `extension_external_ctrl_vars()`; error on unknown `id`.
- `apply_board_update.dock_board` (`R/dock-board.R:343`) — relax the `if (!length(upd$views)) return(board)` early return, then apply the extensions slice.

### 5. Apply + plumbing

`serialize_board.dock_board` already demonstrates that the extension server results (each carrying `$state`) are threadable to the board layer. Thread `ext_res` into the apply path the same way and mirror `apply_block_mod_delta`:

```r
apply_extensions_mod <- function(mod, ext_res) {

  for (ext_id in names(mod)) {

    state_rvs <- ext_res[[ext_id]][["state"]]
    delta <- mod[[ext_id]]

    for (nm in names(delta)) {
      if (!identical(reval_if(state_rvs[[nm]]), delta[[nm]])) {
        state_rvs[[nm]](delta[[nm]])
      }
    }
  }

  invisible()
}
```

`ext_res` is the local already built in `board_server_callback` (`R/board-server.R:65`); make it reachable from the apply path (e.g. via `dock_mgr`, which `apply_board_update.dock_board` already receives through `...`).

## Downstream

- **blockr.md** — implement the contract: promote `content` to a `reactiveVal`, add the state -> editor sync observer, and opt in. Tracked in a companion issue.
- **blockr.assistant** — `tool_list_extensions` / `tool_modify_extension` staging an `extensions$mod` delta, following the existing views tool surface.

## Scope / non-goals

- Field-level `mod` only — no extension `add` / `rm` lifecycle.
- Opt-in; existing extensions (`external_ctrl = FALSE`) untouched.
- dock owns transport + the `reactiveVal` write; each extension owns its controllable vars and its `state -> UI` observer.

nbenn:
## Summary

- Opt-in `external_ctrl` on `new_dock_extension()` (`FALSE` | `TRUE` | character), resolved against the constructor inputs by `extension_external_ctrl_vars()` — the extension analog of blockr.core's `block_external_ctrl_vars()`.
- Controllable `state` entries are enforced to be `reactiveVal`s (read-write) in `validate_ext_srv_result()`; `reval_if` reads them exactly as before, so serialization is unaffected.
- New `extensions$mod` slot on `board_update`, applied through the live extension servers via `dock_mgr$ext_res` (mirrors `apply_block_mod_delta()` — only changed values are written).
- The mod gate (`validate_extensions_delta`) lives in `augment_board_update.dock_board`, the production-reachable validation hook (same place views validate). The `validate_board_update` generic cited in the issue is never invoked by the live board server, so a deep gate there would not fire; it holds only a shallow type-check for the public API, and the deep gate is reached through both the live `preprocess_board_update` path and the generic's `NextMethod()`.
- Non-breaking: extensions default to `external_ctrl = FALSE`. The blockr.md contract (#23) and the assistant tool surface are downstream.

Fixes #162


nbenn:
## Downstream API note (for blockr.assistant#25)

Filed the assistant-side companion: BristolMyersSquibb/blockr.assistant#25 (`list_extensions` / `modify_extension` tools). Implementing it needs one thing from this PR's surface that isn't currently public:

**Export `extension_external_ctrl_vars()`.** It's added here as an internal helper, mirroring core's internal `block_external_ctrl_vars()`. The assistant needs the *resolved* controllable-variable names — to enumerate them in `list_extensions` and to render the "modifiable via modify_extension" hint — and resolving `external_ctrl = TRUE` against the constructor formals is exactly what this helper does. Without an export it would have to call `blockr.dock:::extension_external_ctrl_vars()` or re-implement the `TRUE` → ctor-inputs resolution itself (duplicating dock logic).

I left it internal to match core, but #25 is a concrete caller — so it may be worth promoting it to `@rdname extension` / `@export` before this merges, rather than a follow-up dock release. Happy to add that here if you agree.

**Open question (maybe not this PR):** `list_extensions` would ideally show each controllable var's *current* value. Those live in the per-extension `state` reactiveVals (`dock_mgr$ext_res[[id]]$state`), which the assistant has no handle to. Whether that wants a small dock accessor or is solved assistant-side (operate blind like `modify_block`, which just says "unmentioned keys keep their current values") is worth a thought — flagging it so it isn't lost.

nbenn:
Done — `extension_external_ctrl_vars()` is now exported (`@rdname extension` / `@export`) as of 1af3332, so blockr.assistant#25 can call it directly. The secondary open question (reading each var's *current* value) is left untouched for #25 to weigh.

nbenn:
Resolved the open question above (reading each var's *current* value). As of b560509, every extension server now receives an `extensions` argument — a getter returning the other extensions' server results (keyed by id, each carrying its `$state`). So blockr.assistant#25 can read current values via the server arg, e.g. `isolate(extensions()[["doc_extension"]]$state$content())`, without reaching into internals.

`ext_res` is built lazily by the server `lapply`, so the getter resolves through `dock_mgr` on first read (always post-init, e.g. on a tool call). Additive / non-breaking — existing extensions take `...` and ignore it.

nbenn:
Refined per discussion (a7b7301): `extensions` is now an **environment** with one active binding per id (rather than a getter), so access is natural — `isolate(extensions[["doc_extension"]]$state$content())`, no `()` on `extensions`. The bindings resolve into `dock_mgr$ext_res` (single source of truth, lazy), and ids enumerate via `ls(extensions)` / `names(extensions)`.

PR #168, opened 2026-06-01
Extract grid_map_leaves primitive for layout tree transforms

### LINKED ISSUE #156: Extract a grid_map leaf-transform primitive for the layout tree
## Problem

The dock layout is a recursive grid tree (`leaf` / `branch` nodes). Walking it splits cleanly into two shapes:

- **fold** (read-only accumulate) — already has a shared primitive, `grid_leaves()` (`R/utils-serdes.R`), and `focus_panel()` reuses it. `grid_panel_ids()` predates it and still open-codes, but the primitive exists.
- **map** (rebuild the tree, optionally prune empties) — **no shared primitive.** Each of these open-codes the same `if leaf … else recurse over branch children` dance:
  - `rewrite_grid_leaves()` — `R/layout-class.R` (rewrite leaf view IDs via an id-map)
  - `grid_to_spec()` — `R/utils-serdes.R` (leaf/branch → canonical wire spec)
  - `drop_panels_from_layout()` — `R/views-delta.R` (remove panel IDs from leaves, collapse empty leaves/branches) — added in #150

That's three independent copies of branch recursion on the transform side.

## Proposal

Add a `grid_map` leaf-transform primitive, roughly:

```r
# Apply `fn` to each leaf node; a NULL return prunes the leaf. Branches
# whose children all prune away collapse to NULL too.
grid_map_leaves <- function(grid, fn) { ... }
```

Then route the three map-walkers through it:

- `rewrite_grid_leaves(grid, id_map)` → `grid_map_leaves` with a rename `fn` (never prunes).
- `drop_panels_from_layout()` → `grid_map_leaves` with a `setdiff`-then-prune `fn`.
- `grid_to_spec()` → likely needs a leaf+branch variant (it transforms branches too, not just leaves), so this one may want a more general `grid_map(grid, leaf_fn, branch_fn)`. Worth checking whether the leaf-only form covers enough to be worth it, or whether the general form is the right primitive.

Mirror the fold side: migrate `grid_panel_ids()` onto `grid_leaves()` while we're here.

## Why not in #150

Two of the three map-walkers live in `layout-class.R` / `utils-serdes.R` (the serialization layer), which has its own round-trip test surface. Re-routing them is a cross-cutting refactor that wants its own review, separate from the `views`-slot feature. #150 lands `drop_panels_from_layout` as a direct grid walk; this issue consolidates all three afterwards.

## Done when

- A `grid_map`-style primitive exists with tests (leaf transform + prune + empty-branch collapse).
- `rewrite_grid_leaves`, `drop_panels_from_layout`, and (if it fits) `grid_to_spec` route through it.
- Serialization round-trip tests stay green.

nbenn:
## Summary

- Add `grid_map_leaves(grid, fn)` — the map-side counterpart to the existing `grid_leaves()` fold. It rebuilds the grid, applies `fn` to each leaf, prunes leaves whose `fn` returns `NULL`, and collapses branches that empty out.
- Route `rewrite_grid_leaves()` and `drop_panels_from_layout()` through it, removing their two open-coded branch recursions; fold `grid_panel_ids()` onto `grid_leaves()` to mirror the fold side.
- **Left `grid_to_spec()` standalone** (the issue's "if it fits" walker): it folds the whole tree into a *different shape* — branches become wire `children`/`sizes`, leaves collapse to strings/`panels` — and carries a stricter node contract, so the leaf-only primitive can't express it. Unifying it would take a second general `grid_map(leaf_fn, branch_fn)` whose only non-trivial caller would be `grid_to_spec` itself. Happy to add that if you'd rather see all three consolidated.

Fixes #156

### DIFF
--- a/R/layout-class.R
+++ b/R/layout-class.R
@@ -438,10 +438,8 @@ draw_panel_tree <- function(x) {
 }
 
 grid_panel_ids <- function(grid) {
-
-  views <- unlist(lst_xtr(grid_leaves(grid), "views"), use.names = FALSE)
-
-  if (is.null(views)) character() else views
+  views <- unlst(lst_xtr(grid_leaves(grid), "views"), recursive = TRUE)
+  coal(views, character())
 }
 
 #' @rdname layout-json
--- a/R/layout-class.R
+++ b/R/layout-class.R
@@ -438,8 +438,7 @@ draw_panel_tree <- function(x) {
 }
 
 grid_panel_ids <- function(grid) {
-  views <- unlst(lst_xtr(grid_leaves(grid), "views"), recursive = TRUE)
-  coal(views, character())
+  chr_ply(unlst(lst_xtr(grid_leaves(grid), "views")), identity)
 }
 
 #' @rdname layout-json
--- a/R/layout-class.R
+++ b/R/layout-class.R
@@ -439,23 +439,9 @@ draw_panel_tree <- function(x) {
 
 grid_panel_ids <- function(grid) {
 
-  if (is.null(grid) || !length(grid[["root"]])) {
-    return(character())
-  }
+  views <- unlist(lst_xtr(grid_leaves(grid), "views"), use.names = FALSE)
 
-  walk <- function(node) {
-    if (is.null(node) || !length(node)) {
-      return(NULL)
-    }
-    if (identical(node[["type"]], "leaf")) {
-      return(unlist(node[["data"]][["views"]]))
-    }
-    unlist(lapply(node[["data"]], walk))
-  }
-
-  res <- walk(grid[["root"]])
-
-  if (is.null(res)) character() else res
+  if (is.null(views)) character() else views
 }
 
 #' @rdname layout-json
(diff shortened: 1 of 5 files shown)


R/layout-class.R
```diff
@@ -439,23 +439,9 @@ draw_panel_tree <- function(x) {
...
-      return(NULL)
-    }
-    if (identical(node[["type"]], "leaf")) {
-      return(unlist(node[["data"]][["views"]]))
-    }
-    unlist(lapply(node[["data"]], walk))
-  }
-
-  res <- walk(grid[["root"]])
+  views <- unlist(lst_xtr(grid_leaves(grid), "views"), use.names = FALSE)
```
nbenn: Perhaps `unlst()`? Do you really want recursive flattening?

R/layout-class.R
```diff
@@ -439,23 +439,9 @@ draw_panel_tree <- function(x) {
...
 
-  if (is.null(res)) character() else res
+  if (is.null(views)) character() else views
```
nbenn: This calls for `coal()`, no?

R/views-delta.R
```diff
@@ -239,46 +239,29 @@ drop_panels_from_layout <- function(layout, panel_ids) {
...
-
-    if (identical(node[["type"]], "leaf")) {
-
-      views <- unlist(node[["data"]][["views"]], use.names = FALSE)
-      kept <- setdiff(views, panel_ids)
-
-      if (length(kept) == 0L) {
-        return(NULL)
-      }
+    views <- unlist(leaf[["data"]][["views"]], use.names = FALSE)
```
nbenn: See earlier comment.

nbenn (commented):
Lgtm.

R/layout-class.R
nbenn: Switched to `unlst()`. And yes, recursive is intended here: `lst_xtr(grid_leaves(grid), "views")` yields a list of per-leaf `views`, each of which is itself a list, so the IDs sit two levels deep. Made the intent explicit with `recursive = TRUE`.

R/layout-class.R
nbenn: Done — now `coal(views, character())`.

R/views-delta.R
nbenn: Switched to `unlst()` here too. This one is a single leaf's `views` (one level), so the default `recursive = FALSE` is correct — no flag needed, unlike `grid_panel_ids()`.

R/layout-class.R
nbenn: Good point — `recursive = TRUE` would mask an unexpected extra level. Switched to handling the known two levels explicitly: `chr_ply(unlst(lst_xtr(grid_leaves(grid), "views")), identity)`. `unlst()` peels the leaf grouping (one level) and `chr_ply(…, identity)` coerces each leaf's views, erroring on anything deeper instead of silently flattening it (cc847f5).

R/layout-class.R
nbenn: Follow-up: with the `chr_ply()` form, the empty case yields `character(0)` rather than `NULL`, so the `coal()` guard became redundant and I dropped it. Happy to restore it if you'd prefer it explicit.

Issue #166, opened 2026-06-01
Decouple view identity from name: give views stable ids


nbenn:
## Summary

Views are keyed by their user-editable **name** everywhere — there is no separate immutable id. This couples display label and identity, forces naming constraints, makes rename a re-key across several structures, and blocks the cleanest refactor direction discussed in #164 (reconcile-against-board-state). This issue proposes giving each view a stable id, decoupled from its name — the same id/name separation already used for blocks, links, etc. (e.g. `block_name()` is an editable label over an immutable block id).

## Current state: name *is* identity

A `dock_layouts` is a named list whose **names are the view names**; identity lives in the name:

- The active view is an `active = TRUE` attribute on a layout element, and `active_view()` returns that element's **name** (`R/layouts-class.R:141-150`).
- The runtime registry `dock_mgr$docks` is **also** keyed by view name (`R/board-server.R:204`, `:270`).
- Names are constrained precisely because they're keys: unique + non-empty at the model level (`validate_dock_layouts`, `R/layouts-class.R:89-94`) and `[a-zA-Z0-9 _-]` + unique at the UI level (`validate_view_name`, `R/board-server.R:1149`).

There *is* a stable `dock_id` (a random `dock_xxx`), but it only namespaces the DOM element and the Shiny module — it is **not** the key into the persisted `dock_layouts` or the live `docks` registry. That is why `rename_view_observer` has to re-key both structures in place, deliberately keeping `dock_id` so the live module and DOM survive (`R/board-server.R:949-952`).

## Why this matters

1. **Rename is a multi-structure re-key** rather than a single attribute write: the persisted layouts, `vs$state`, `dock_mgr$docks`, and the nav tabs must all be updated in lockstep by name.
2. **Naming constraints exist only because names are keys.** With a stable id, uniqueness / charset become pure display concerns (or disappear).
3. **It blocks the reconcile-against-board-state design in #164.** That approach diffs `board_layouts` (name-keyed) against the live registry (name-keyed); with no id to correlate "A → B", a rename is indistinguishable from remove-A + add-B and would tear down and rebuild the live dock. A stable id lets the reconciler match views across renames.

## Proposed direction

Adopt the same id/name separation used elsewhere in blockr: an **immutable view id** as the structural key, with the **view name** as a separate, freely editable display attribute. Details for the implementing agent to settle:

[...]

PR #170, opened 2026-06-01
Pass dock manager down via board_server() instead of callback return

### LINKED ISSUE #164: dock_manager is a threaded mutable handle forced by side effects in apply_board_update.dock_board
## Summary

`new_dock_manager()` (`R/board-server.R:102`) builds a bare environment that is
passed around the board-server code as if it were a proper object: it mimics the
`new_*` constructor convention but carries no class, no validator, and no single
declaration of its fields. Its shape is assembled in three places — the
constructor sets `docks`/`active_dock`/`next_id`; `board_server_callback` bolts
on `board_rv`/`update`/`triggers`/`vs` (`R/board-server.R:37-51`);
`init_view_docks` bolts on `current_active` (`R/board-server.R:181`).

This issue argues that the environment — and most of its members — is **not an
inherent requirement** but an artifact of one placement decision, and proposes a
refactor that largely dissolves the "manager" rather than promoting it to a
class.

## What the manager actually is

A server-side **mutable handle threaded across the package boundary**. It is
created in `board_server_callback`, returned in the callback result as
`dock_mgr = dock_mgr` (`R/board-server.R:79`), and read back out in
`apply_board_update.dock_board` via `dock_mgr <- dot_args$dock_mgr`
(`R/dock-board.R:351`). The thread-back works because blockr.core merges the
callback's return value into `dot_args` (`blockr.core/R/board-server.R:104`,
`dot_args <- c(dot_args, cb_res)`) and forwards it as `...` to
`apply_board_update`.

It is mutated from **two independent reactive contexts**:

- the UI observers in `board_server_callback` (`switch_/add_/remove_/rename_view_observer`), and
- the `apply_views_{rm,add,mod,active}()` functions called from the S3 method `apply_board_update.dock_board`, which runs inside core's update observer — outside the callback closure.

Because both mutate the same `docks`/active-view state, the handle must be
shared **by reference** → environment.

## Root cause

The forcing decision is that **`apply_board_update.dock_board` performs live UI
surgery** (`insertUI`, `manage_dock`, `destroy_module`, `sendCustomMessage`;
see `apply_views_add`, `R/views-delta.R:410-439`). In blockr.core,
`apply_board_update` is a **pure `(board, upd) -> board` reducer** — the default
method is a no-op:

```r
# blockr.core/R/board-server.R:1192
apply_board_update.board <- function(board, upd, ..., session = get_session()) {
  board
}
```

Core's equivalent side effects (`insert_block_ui`, `setup_block`, link rewiring)
do **not** live in this hook. They live in a separate privileged function,
`apply_core_board_update(rv, upd, …)` (`blockr.core/R/board-server.R:1197`),
which runs first (`:177`), has the live `rv` reactiveValues in scope, and
mutates it in place. The pure hook runs afterwards (`:187`) only to compute a
new board *value*.

blockr.dock put view side effects in the reducer hook instead of in a
closure-resident observer. That method is dispatched lexically **outside**
`board_server_callback`, so it cannot see dock's live per-session state — hence
the handle has to be returned and threaded back, and hence it has to be an
environment.

## Members: load-bearing vs. artifact

| member | reason it's on the env | needs the env? |
|---|---|---|
| `docks` | registry of live `manage_dock()` module results keyed by view | genuine mutable identity-bearing state — but can be closure-private |
| `next_id` | unique dock-id generator | no — already just a closure |
| `active_dock`, `vs`, `current_active` | reactiveValues / reactiveVal | no — already reference types; on the env only because it became the carry-all |
| `board_rv`, `update`, `triggers` | so `apply_views_add` can call `manage_dock` out-of-closure (`R/views-delta.R:429-431`) | no — pure stashing artifact; already in `board_server_callback` scope |

A small mutable registry of live dock modules is intrinsic (we track running
Shiny modules by view name). Everything else is downstream of the placement
decision.

## Corroborating symptom: view-CRUD implemented twice

The same `insertUI`/`manage_dock`/`destroy_module` surgery exists in two places,
triggered by two entry points:

- UI-driven: `add_view_observer` / `remove_view_observer` / `rename_view_observer` (`R/board-server.R`)
- delta-driven: `apply_views_{add,rm,mod,active}` (`R/views-delta.R`)

The delta-driven copy arrived with #150 (`3a4138c`, structured `views` slot) and
is what bolted `board_rv`/`update`/`triggers` onto the manager so the new path
could start modules. #159 (orphaned nav tab on `apply_views_rm`) is the kind of
drift this duplication produces.

## Proposed direction

Adopt core's own separation:

1. Make `apply_board_update.dock_board` **pure** — update `board_layouts(board)`
   and return `board`; no `dock_mgr`, no session side effects.
2. Move the UI surgery into an observer that lives **inside**
   `board_server_callback`, driven either by reconciling against the updated
   `board_layouts(board$board)` or by observing the `update()` signal's `$views`
   slot. Such an observer already closes over `docks`, `session`, `board`,
   `update`, `triggers` — nothing needs to leave the function.
3. Collapse the duplicated UI-driven and delta-driven CRUD into that single
   reconciliation path (single source of truth = board layouts; single mutator).

What survives is a small **private** `docks` registry mutated from one place.
`board_rv`/`update`/`triggers` stop being fields; `active_dock`/`vs`/
`current_active` go back to being closure locals. No threaded object, no
`new_dock_manager()` masquerade. If a residual registry still benefits from a
name, it can be a minimal classed S3 value rather than a catch-all environment.

## Open questions / caveats

- The current design's one real advantage is **synchronous** application: it has
  the delta in hand and applies it without re-diffing or an extra reactive
  flush. A pure reducer + reconciliation observer trades one flush for the
  simplification; the `update()`-observing variant keeps it synchronous *and*
  in-closure, so the advantage looks recoverable — needs confirmation.
- Decision to make: reconcile-against-board-state vs. observe-the-`update()`-payload.
- Interacts with #152 (rename `dock` → `active_dock` extension arg) — the
  `active_dock` reactiveValues mirror is part of this surface.

## Acceptance / scope

- [ ] `apply_board_update.dock_board` and `apply_views_*` no longer perform DOM /
  module side effects (pure board-value transforms).
- [ ] View add/remove/rename/switch handled by a single closure-resident path.
- [ ] `new_dock_manager()` removed or reduced to a minimal private registry; no
  cross-boundary threading via `dot_args$dock_mgr`.
- [ ] Tests in `test-board-server.R` / `test-views-delta.R` updated; behaviour
  (multi-view add/remove/rename/switch, external `views` deltas) preserved.

_Structural analysis only — not yet validated against a running session._

nbenn:
## Summary

- Create the per-session dock manager at the entry point (`blockr_app_server.dock_board()`) and thread it **down** to `board_server_callback()` and `apply_board_update.dock_board()` through `board_server()`'s `...` / `dot_args`, instead of creating it in the callback and threading it **back** via core's callback-return-to-`dot_args` merge. Removes the handle's round-trip and the constructor masquerading as a callback return value.
- `apply_board_update.dock_board()` is unchanged — it still reads `dot_args$dock_mgr`, now populated by the down-thread.
- `new_dock_manager()` is now a classed `dock_manager` with a validator.
- Session-scoped reactives (`active_dock`, …) are attached in the callback, not at the entry point, so they bind to the board module's reactive domain rather than the outer app session — creating them outside the module caused a view-update loop (caught by the `shinytest2` app test).

Scope: side effects stay in `apply_board_update.dock_board()` (the designed post-apply hook — see issue discussion); view stable-ids and rename handling are split out to #166.

Fixes #164

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # blockr.dock (development version)
 
+* The per-session dock manager is created at the board-server entry
+  point (`blockr_app_server.dock_board()`) and passed *down* to
+  `board_server_callback()` and `apply_board_update.dock_board()`
+  through `board_server()`'s `...` / `dot_args`, instead of being
+  created in the callback and threaded *back* through the merged
+  callback return. It is now a classed `dock_manager` with a
+  validator. Internal refactor; no change to board behaviour (#164).
+
 * The add / append / prepend block action handlers now mount the
   `blockr.ui` block-browser module: a card-list block picker with
   search, per-category icons, and a per-card expand for tweaking the
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -8,13 +8,21 @@
 #' @param board Reactive board state (list with `$board`).
 #' @param update Reactive update signal from blockr.core.
 #' @param ... Extension server arguments.
+#' @param dock_mgr Per-session dock handle from [new_dock_manager()], created
+#'   at the board-server entry point and threaded in through `board_server()`'s
+#'   `...` so the apply path reaches the live dock registry.
 #' @param session Shiny session.
 #'
 #' @return List with `dock`, `actions`, `view_data`, and extension
 #'   results.
 #'
 #' @noRd
-board_server_callback <- function(board, update, ..., session = get_session()) {
+board_server_callback <- function(board, update, ...,
+                                  dock_mgr = new_dock_manager(),
+                                  session = get_session()) {
+
+  validate_dock_manager(dock_mgr)
+
   initial_board <- isolate(board$board)
 
   c_blks <- board_blocks(initial_board)
@@ -33,7 +41,7 @@ board_server_callback <- function(board, update, ..., session = get_session()) {
 
   views <- board_layouts(initial_board)
 
-  dock_mgr <- new_dock_manager()
+  dock_mgr$active_dock <- reactiveValues()
   dock_mgr$board_rv <- board
   dock_mgr$update <- update
   dock_mgr$triggers <- triggers
@@ -102,8 +110,7 @@ board_server_callback <- function(board, update, ..., session = get_session()) {
     list(
       dock = dock,
       actions = triggers,
-      view_data = view_data,
-      dock_mgr = dock_mgr
+      view_data = view_data
     ),
     ext_res
   )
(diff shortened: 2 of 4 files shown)


nbenn:
Superseded by #176, which stacks the #164 work on the stable-view-id base (#175 / `174-container-active`). Rather than the cosmetic de-thread here, #176 makes `apply_board_update.dock_board` a pure reducer and dissolves the threaded handle via a closure-resident reconcile pass — the dramatic simplification the stable ids unlock.

PR #172, opened 2026-06-02
feat: adopt stack_menu module in add/edit stack actions

DivadNojnarg:
## Summary

Phase 2 of the [stack-menu openspec change](https://github.com/BristolMyersSquibb/blockr.ui/tree/add-stack-menu/openspec/changes/add-stack-menu). \`add_stack_action\` and \`edit_stack_action\` now mount the new \`blockr.ui::stack_menu_ui()\` / \`stack_menu_server()\` module and observe a single committed-stack reactive.

### Changes

- **per-field inputs removed**: \`stack_id\`, \`stack_name\`, \`stack_color\`, \`stack_block_selection\`, \`stack_confirm\` and their \`edit_stack_*\` equivalents are gone. The handlers receive everything via the menu's committed reactive.
- **\`shinyWidgets::colorPickr\` dropped**: the new menu's inline hue + lightness slider colour picker replaces it (no popover escaping the sidebar).
- **\`stack_sidebar_body()\` deleted** from \`R/action-sidebar.R\` (no in-tree callers remain; \`link_sidebar_body()\` stays).
- **Edit-mode title reads \`Edit stack <id>\`** so the user sees which stack they're editing - the menu body no longer prints its own subtitle.
- **\`mod\` entries are partial-arg deltas**: \`list(blocks, name, color)\` keyed by stack id, matching blockr.core's new \`update_stack()\` contract. The old \`as_stacks()\` wrapping trips \`board_update_stacks_mod_entry_invalid\` under the new schema.

### Out of scope

The \`Remotes\` pin to \`blockr.ui@add-stack-menu\` reverts to the bare remote once [blockr.ui#9](https://github.com/BristolMyersSquibb/blockr.ui/pull/9) merges.

## Test plan

- [x] \`devtools::test()\` on \`tests/testthat/test-action-stack.R\` - 26/26 pass (5 tests: create valid, create invalid id / name / colour, edit valid, edit invalid name / colour, remove).
- [x] \`lintr::lint_package()\` clean under the CI config.
- [ ] Reviewer: manual smoke in the dock + dag example - right-click empty area / open stacks panel -> Create new stack -> pick 2-3 blocks via cards, type name, confirm; right-click an existing stack -> Edit -> deselect a block, change name + colour via the inline picker, Update.


DivadNojnarg:
Test app with `blockr.ui@add-stack-menu`:

```r
library(blockr.core)
library(blockr.dock)

board <- new_dock_board(
  extensions = blockr.dag::new_dag_extension(),
  blocks = c(
    dataset_1 = new_dataset_block(),
    head_1 = new_head_block()
  ),
  links = new_link("dataset_1", "head_1"),
  layouts = list(
    Analysis = dock_layout(
      "dag_extension",
      panels("dataset_1", "head_1", active = "head_1"),
      sizes = c(0.3, 0.7)
    ),
    Overview = dock_layout("dag_extension", active = TRUE),
    Empty = list()
  )
)

serve(board, "my_board")
```

1. Right click on DAG.
2. Click "create stack".
3. Choose options and validate.
4. Right click on the new stack.
5. Click "Edit stack".
6. Make changes.
7. Validate.

R/action-stack.R
```diff
@@ -1,96 +1,52 @@
...
+      }
+
+      observeEvent(trigger(), {
+        blockr.ui::show_sidebar(
+          sidebar_id, title = "Create new stack", ui = menu_ui()
+        )
+      })
+
+      observeEvent(committed(), {
+        spec <- committed()
```
nbenn: We could simplify the code here even further, no? What if `blockr.ui::stack_menu_server()` returned a `stacks` object directly (containing `dock_stack` objects)?

R/action-stack.R
```diff
@@ -1,96 +1,52 @@
...
+
+        if (!valid_stack_id(spec$id, board$board, session)) return()
```
nbenn: The validators could also live in blockr.ui. You could pass `board` and `session` to the stack menu server too.

DivadNojnarg: @nbenn Sure. There was an important gap anyway because the server module was not considering the board, and was thereby unable to update the displayed blocks in the sidebar menu. Also spotted other issues on the way, which are solved now (edge case when removing a stack with the pinned sidebar in edit mode which would cause a crash ...)

R/action-stack.R
DivadNojnarg: @nbenn I did something in block.ui to use the blockr.core stacks API. This however required some changes here in dock-stack.R in `as_dock_stack.stack` method (color was lost at some point).

R/action-stack.R
```diff
@@ -1,96 +1,47 @@
...
+        blockr.ui::show_sidebar(
+          sidebar_id, title = "Create new stack", ui = menu_ui()
+        )
+      })
+
+      observeEvent(committed(), {
+        # The menu returns a `blockr.core` stacks object (id-keyed, colour
+        # carried as an attribute); `as_dock_stacks()` re-classes it to
+        # `dock_stack` preserving that colour. No reshaping here.
+        update(list(stacks = list(add = as_dock_stacks(committed()))))
```
nbenn: Why not return a `stacks` object consisting only of `dock_stack` objects directly form blockr.ui?

R/dock-stack.R
```diff
@@ -38,7 +38,7 @@ validate_stack.dock_stack <- function(x) {
 
   col <- stack_color(NextMethod())
 
-  if (!is_string(col) || !is_hex_color(col)) {
+  if (!is_string(col) || !blockr.ui::is_hex_color(col)) {
```
nbenn: I don't think we should move `is_hex_color()` to blockr.ui. How `dock_stack` colors are represented is a concern of blockr.dock.

R/dock-stack.R
```diff
@@ -129,7 +129,9 @@ as_dock_stack.dock_stack <- function(x, ...) {
 }
 
 #' @export
-as_dock_stack.stack <- function(x, color = suggest_new_colors(), ...) {
+as_dock_stack.stack <- function(x,
+                                color = attr(x, "color") %||%
```
nbenn: This feels wrong to me. A vanilla `stack` object should not have a color attribute. We have `dock_stack` for this. If you create a vanilla `stack` object first, but must attach a color attribute, something is off. Better address the root cause.

R/dock-stack.R
```diff
@@ -173,13 +175,26 @@ as_dock_stacks <- function(x) {
...
   }
 
-  new_col <- suggest_new_colors(
-    chr_ply(x[!todo], stack_color),
-    n = sum(todo)
+  # A non-dock stack may already carry a `color` attribute (e.g. one
+  # built by `blockr.ui`'s stack menu via `new_stack(..., color = )`):
+  # keep it. Only stacks lacking a colour get a freshly suggested one,
+  # seeded against the existing palette + the carried colours.
+  carried <- chr_ply(
```
nbenn: None of this should be necessary. If we need to coerce stack objects to `dock_stack` we generate colors, b/c they should not already carry colors. Again this this feels like a workaround for something.

nbenn (commented):
There are some issues around instantiations of `dock_stack` objects and I feel you're trying to work around a circular dependency issue with this. The solution imo should not be creating `stack` objects that already carry color attributes and then abusing coercion logic here to deal with such malformed objects, but to solve the dependency issue with making it a suggests dep in one direction.

How the dependency structure *should* look like, I think is guided by conceptual considerations: does blockr.ui make sense without blockr.dock? I.e. could one use blockr.ui with blockr.core? From a distance, I think what we have right now is in some sense the wrong way round: blockr.ui *augments* blockr.dock, but should *not* be a strict requirement for running blockr.dock. Ideally, from a conceptual perspective, blockr.dock would run fine without blockr.ui, albeit with a degraded ux. In that sense, blockr.ui should be a suggests dep of blockr.dock and blockr.dock should be an imports dep of blockr.ui.

What do you think about this? If we want to go for this, we could make the actions non-functional unless blockr.ui is available. The `blockr` meta-pkg can always have both blockr.ui *and* blockr.dock a hard deps.

R/action-stack.R
DivadNojnarg: Done. blockr.ui now builds the `dock_stack`s itself via `blockr.dock::new_dock_stack()`, gated on `pkg_avail("blockr.dock")` (a Suggests back-edge, so no import cycle). The handlers apply `committed()` as-is, the `as_dock_stacks()` wrap is gone. blockr.ui 3281a35, blockr.dock ed6d2b8.

R/dock-stack.R
DivadNojnarg: Agreed. `is_hex_color()` is back in blockr.dock and the `blockr.ui::` calls are dropped. blockr.ui keeps its own copy for validating the menu input.

R/dock-stack.R
DivadNojnarg: Fixed at the source: the menu returns `dock_stack`s now, so nothing hangs a colour on a vanilla `stack`. `as_dock_stack.stack` is back to its main version.

R/dock-stack.R
DivadNojnarg: Gone. `as_dock_stacks()` is reverted to main: bare stacks just get a fresh colour, no carried-colour handling.

DivadNojnarg:
As discussed, for this PR, to be pragmatic, we mark blockr.dock as a suggest of blockr.ui + adding some guards if blockr.dock isn't available. Long term, see this follow up issue: https://github.com/BristolMyersSquibb/blockr.ui/issues/12

PR #173, opened 2026-06-02
Give views stable ids decoupled from their display name

### LINKED ISSUE #166: Decouple view identity from name: give views stable ids

nbenn:
## Summary

- Views now carry an immutable **id** (the key into `dock_layouts` and the `dock_mgr$docks` registry) separate from the editable display **name**, mirroring the block id/name split. `active_view()` returns the id; `view_name()` / `view_name<-()` / `view_names()` read and write the label.
- Rename collapses to a name-attribute write: the id — and therefore the dock module and DOM element — is untouched, so nothing is re-keyed and the live-sync rename no longer leaks as remove-then-add. Dock module / DOM ids derive deterministically from the view id (no random per-render minting), and the `views` delta gains a cheap `rename` slot.
- Naming constraints relax to display concerns (non-empty, unique label); the charset restriction is gone.
- Serialization round-trips ids and migrates id-less boards by minting ids on load.
- Name-keyed `views` deltas are resolved to ids at ingest, so name-based producers (e.g. `blockr.assistant`) keep working unchanged.
- Keystone for the reconcile-against-board-state refactor discussed in #164.

## Migration

| Before | After |
| --- | --- |
| `active_view(x)` returns a view name | `active_view(x)` returns a view **id** |
| `board_layouts(x)[["My View"]]` | `board_layouts(x)[[<id>]]` — resolve a name with `view_names(x)` |
| name is identity (unique, `[a-zA-Z0-9 _-]`) | id is identity; name is a free-form, unique display label |
| serialized views keyed by name | keyed by id, name carried alongside (old boards migrate on load) |

Fixes #166


R/layouts-class.R
```diff
@@ -30,36 +46,87 @@
...
-    inner <- vws[[1L]]
-    inner_nms <- names(inner)
-    if (!is.null(inner_nms) && all(nzchar(inner_nms))) {
-      vws <- inner
-    }
+    vws <- vws[[1L]]
   }
 
   if (!length(vws)) {
     vws <- list(Page = new_dock_layout())
```
nbenn: This does not look like a good default. The ID should be generated via ID generation mechanism and the name then derived.

R/layouts-class.R
```diff
@@ -30,36 +46,87 @@
...
+    ids[!given] <- rand_names(c(reserved, ids[given]), n = sum(!given))
+  }
+
+  set_names(views, ids)
+}
+
+# View ids key the container, so they must be DOM- and namespace-safe
+# identifiers — free-form display labels live on the object as the name.
+validate_view_ids <- function(ids) {
+
```
nbenn: Should we not check for uniqueness here too? Or should we fold this into what `validate_dock_layouts()` already does?

R/layouts-class.R
```diff
@@ -131,6 +198,52 @@ layout_ids <- function(x) {
...
 }
 
+#' @param x A `dock_layout` (for `view_name()` / `view_name<-()`) or a
+#'   `dock_layouts` collection (for `view_names()`).
+#' @rdname view
+#' @export
+view_name <- function(x) {
+  # exact = TRUE: a bare "name" would otherwise partial-match the "names"
+  # attribute and return the layout's element names.
+  attr(x, "name", exact = TRUE)
```
nbenn: Would `view_name` be a better attribute name? Given how R happily partial matches things, this is a bit of a footgun.

R/layouts-class.R
```diff
@@ -131,6 +198,52 @@ layout_ids <- function(x) {
...
+
+# A view's display label: its explicit name, or — when unset — one
+# derived from the id, the same way blockr.core derives a default block
+# name from its class (underscores to spaces, capitalise the first
+# letter). The id is the collection key, so this needs both.
+view_label <- function(layout, id) {
+  coal(view_name(layout), humanize_id(id), fail_all = FALSE)
+}
+
+humanize_id <- function(id) {
```
nbenn: How about a better name for this? Something like `name_from_id()` or so?

R/layouts-class.R
```diff
@@ -131,6 +198,52 @@ layout_ids <- function(x) {
...
+  res <- gsub("_", " ", id)
+  paste0(toupper(substr(res, 1L, 1L)), substring(res, 2L))
+}
+
+# Resolve a display label to its view id within a collection. Returns the
+# first match (display labels are kept unique at the input / UI boundary)
+# or `NA`.
+view_id_by_name <- function(x, name) {
```
nbenn: Given that names are non-unique, the fact that we have this function worries me. What is this used for? I'd prefer if we removed it.

R/views-delta.R
```diff
@@ -40,75 +40,155 @@ resolve_views_layouts <- function(views, board, upd) {
...
+    return(names(added)[added_hit])
+  }
+
+  ref
+}
+
+# Wire-added views are keyed by the desired display name (the LLM / UI
+# supplies a name, not an id): mint a fresh id for each — unique against
+# the board's existing ids — and carry the key over as the display name.
+mint_added_view_ids <- function(add, reserved) {
```
nbenn, line 60: This feels wrong. The UI (e.g. a selectize with a mapping name/id) should key by ID. The name is just sugar for the user. An LLM should also interact *only* interact via IDs. This is the only unique and stable handle we have.

nbenn (commented):
See comments.

PR #175, opened 2026-06-02
Track the active view on the dock_layouts container

### LINKED ISSUE #174: Track the active view on the container, not as a per-view attribute
## Summary

A `dock_board` can hold several views but only one is shown at a time — the *active* view. Today "active" is stored on each view as an attribute on the `dock_layout` itself, so the "exactly one active at a time" invariant is spread across N independent booleans on N independent objects. Nothing in the representation prevents zero, or two-plus, views from being active; the invariant is only *recovered* after the fact by scanning, counting, and normalizing. This issue proposes moving the active marker onto the container (`dock_layouts`) as a single field that references the active view, so the invariant holds by construction — the container is the only place it can actually be enforced.

## Current model

`set_active_view()` writes a per-view attribute and `is_active_view()` reads it (`R/layouts-class.R:226-238`):

```r
is_active_view <- function(x) {
  is_dock_layout(x) && isTRUE(attr(x, "active"))
}

set_active_view <- function(x, active = TRUE) {
  stopifnot(is_dock_layout(x))
  attr(x, "active") <- if (isTRUE(active)) TRUE else NULL
  x
}
```

Because the truth is distributed across the views, every operation has to reconstruct or repair the single-active invariant rather than rely on it:

- **Read** — `active_view.dock_layouts()` scans all views and takes the *first* one whose flag is set, silently ignoring any others (`R/layouts-class.R:141-150`).
- **Write** — `active_view<-.dock_layouts()` loops over *every* view to clear the flag on the others (`R/layouts-class.R:165-180`); a partial write would leave multiple views active.
- **Construct** — `new_dock_layouts()` patches the "zero active" case by force-activating the first view (`R/layouts-class.R:56-58`).
- **Validate** — `validate_dock_layouts()` catches the "two-or-more active" case only as a post-hoc count check (`R/layouts-class.R:105-112`).
- **Serialize** — the marker is emitted twice on the container path: once at the container level via `active_view(lys)` (`R/utils-serdes.R:60`) and again as a redundant per-view boolean inside each serialized view via `is_active_view()` (`R/utils-serdes.R:50`). On load, `blockr_deser.dock_layouts()` ignores the per-view booleans entirely and trusts the container field (`R/utils-serdes.R:106`) — direct evidence that the per-view flag is dead weight inside a container.

## Proposed change

Make the active view a single field on `dock_layouts` (the container) holding the active view's **id**, and drop the per-view `active` attribute as the source of truth. The container then owns the invariant outright:

- "exactly one active" becomes structural — one field, one value — instead of something reconstructed by scan / count / normalize on every access;
- the writer is a single assignment, not an N-way loop that can half-finish;
- construction defaults the field to the first view, and an unknown id is rejected at the boundary;
- serialization carries the marker exactly once, at the container level.

The public accessors `active_view()` / `active_view<-()` already abstract the representation (`R/layouts-class.R:136-188`), so most callers are unaffected and the change stays behind the existing interface. What gets rewritten or removed are the representation-specific internals — `is_active_view()` / `set_active_view()` / `any_active_view()` — and their call sites across `R/dock-board.R`, `R/views-delta.R`, `R/utils-serdes.R`, and `R/layout-class.R`.

## Blocked by #173

A container-level field has to point at something stable. View **names** are user-editable (and, before #173, were themselves the identity), so the field must reference the immutable view **id** introduced in #173, where `active_view()` already returns the id rather than the name. Building this on names would re-introduce exactly the rename-as-rekey problem that #173 removes. This refactor should therefore rebase onto #173, whose serialization migration is also the natural place to fold the old per-view flag into the new container-level field.

## Out of scope / related

The *runtime* mirror of the active view — `dock_mgr$current_active` and the swapped `dock_mgr$active_dock` reactiveValues (`R/board-server.R`) — is a separate representation tracked by #164. This issue is about the serializable state model (`dock_layouts`); keeping that model and the runtime handle consistent is part of #164's reconcile-against-board-state work.

nbenn:
## Summary

- Active view is now a single field on the `dock_layouts` container (the active view's id) rather than a per-view `active` attribute, so the "exactly one active" invariant holds by construction instead of being recovered by scan / count / normalize on every access.
- Drops `is_active_view()` / `set_active_view()` / `any_active_view()`. `active_view()` / `active_view<-()` read and write the one container field, `validate_dock_layouts()` no longer needs a multiple-active count check, and serialization emits the marker once at the container level — the redundant per-view boolean is gone from the wire format.
- Adds "add a view and activate it in one delta": a views delta's `active` may forward-reference a view being added, by its `add` key, resolved to the freshly minted id in `normalize_views_delta()` before validation runs. This is the public, id-free mechanism requested in the issue comment (for BristolMyersSquibb/blockr.assistant#30); the New page modal now uses it in place of the retired per-view marker.

`dock_layout(active = TRUE)` still marks a view at construction — it is now a transient hint folded into the container field. Per-view `active` booleans in older saved layouts are ignored on load; the container `active` id is authoritative.

Stacked on #173 — base is `166-view-ids`. Retarget to `main` once #173 lands.

Fixes #174


R/layouts-class.R
```diff
@@ -322,19 +314,15 @@ as_dock_layouts.dock_layouts <- function(x, ...) x
...
-  attr(x, "active") <- if (isTRUE(active)) TRUE else NULL
-  x
+# Wrap an id-keyed list of views as a `dock_layouts`, recording the active
+# view as a single field on the container. Active defaults to the first
+# view; callers choose a different one through `active_view<-()` (or, at
+# board construction, `new_dock_board(active = )`). A view never carries
+# its own active marker — the container is the sole owner.
+finalize_layouts_active <- function(views) {
+  active <- if (length(views)) names(views)[1L] else NULL
+  structure(views, class = "dock_layouts", active = active)
```
nbenn: Should we not thread this through a constructor/coercion function call?

R/dock-board.R
```diff
@@ -58,63 +61,66 @@ new_dock_board <- function(blocks = list(), links = list(), stacks = list(),
...
+    active_view(res) <- active
+  }
+
+  res
+}
+
+# A multi-view input is a list keyed by view id, or a list whose elements
+# are themselves `dock_layout`s (keyless — ids minted). A bare list of
+# panel ids (`list("a", "b")`) is a single view's children.
+is_multi_view <- function(layouts) {
```
nbenn: Not sure about introducing the "multi-view" concept like this. it simplifies branching conditions a bit, but tbh, I'd rather we use coercion functions `as_dock_layouts()`/`as_dock_layout()` to homogenize inputs before further processing, instead of such branching logic.

nbenn (commented):
See comments.

PR #183, opened 2026-06-03
feat: adopt link-menu module in add_link_action (Phase 2)

DivadNojnarg:
Phase 2 of the `add-link-menu` OpenSpec change: `blockr.dock` adopts the
`blockr.ui` link-menu module.

> **Stacked PR** — based on `add-stack-menu-adoption` (not yet merged).
> Retarget to `main` once that PR lands so the diff stays scoped to the
> link-menu changes.

## What changed

- **`add_link_action()`** rewritten to mirror the stack-menu adoption:
  mounts `blockr.ui::link_menu_server("menu")` once and renders
  `link_menu_ui()` via a small `menu_ui()` helper.
- **Bidirectional**: the JS binding composes `source` / `target` from
  the clicked card's `data-direction` + the panel anchor, so
  right-clicking a downstream block now lets the user pick an upstream
  source — not just a target.
- **Single committed observer** validates the link id, resolves the
  target port (explicit pick → first free slot → fresh variadic slot via
  `block_input_select(mode = "inputs")`), and adds the link.
- **Multi-link session**: after each commit, pools are recomputed with
  `link_eligible_pools()` on the next flush and pushed via a
  `pool-update` `sendInputMessage`, so the just-wired card drops
  client-side without a re-render; the sidebar closes itself once both
  pools drain. The pool-update payload is shaped
  (`eligible = list(outgoing, incoming)` + top-level `free_inputs`) to
  match what the JS `receiveMessage` reads.
- The menu owns its empty-state, so the pre-flight `NULL`-check /
  notification is removed.
- **Removed** `link_sidebar_body()` / `R/action-sidebar.R` (no callers
  remain) and bumped the `blockr.ui` Remote to `@add-link-menu`.

## Tests

`tests/testthat/test-action-link.R` rewritten to drive `menu-commit`
directly: OUTGOING default port, OUTGOING explicit port (arity-2 merge),
INCOMING (anchor as target), duplicate-id rejection, empty-pool still
opens the sidebar. Full suite: 0 failures.

## Checks

`devtools::check()` clean under a UTF-8 locale: 0 errors / 0 warnings /
2 benign NOTEs (timestamp + local `CLAUDE.local.md`).


DivadNojnarg:
Merge after #172

R/action-link.R
```diff
@@ -4,128 +4,83 @@ add_link_action <- function(trigger, board, update, ...) {
...
+
+        links <- board_links(board$board)
+        trg_blk <- board_blocks(board$board)[[spec$target]]
+
+        # The menu only renders a port picker for finite-arity targets
+        # with > 1 free slot, so `spec$block_input` arrives NULL for
+        # arity-1 and variadic targets. Resolve the slot ourselves in
+        # those cases: the first free named input, or - for variadic
+        # targets - a freshly generated slot, both via the same helper.
+        chosen_input <- spec$block_input
```
nbenn: Why not move all of this into blockr.ui? So that the server here just returns a ready-made `links` object?

nbenn (commented):
Lgtm. The only comment i have if more of a minor tweak. Nothing wrong with what we have atm. At the same time, it might be a bit cleaner to have blockr.ui just return an object that is ready to be consumed without further pre-processing. I'm not even sure that this is needed. It might be that blockr.core correctly sorts out missing `input` entries. But I'm not entirely sure it does. You'd have to check that.

R/action-link.R
DivadNojnarg: Done. `link_menu_server()` resolves the target port and returns a ready-to-apply `links` object now, so the handler just does `update(list(links = list(add = committed())))`. All blockr.core under the hood (no `dock_link` type), so no Suggests gate needed unlike the stack menu. blockr.ui 09ba4d5, blockr.dock 6a884d3.

PR #201, opened 2026-06-04
Report on-screen blocks to core's visibility channel

### LINKED ISSUE #195: Report on-screen blocks to core's visibility channel
blockr.core can gate block evaluation and rendering on visibility (BristolMyersSquibb/blockr.core#203), but core has no way to know what is on screen. blockr.dock is the only component that does. This implements the producer side.

## Approach

In `board_server_callback()`, compute the set of on-screen block ids and write it to core's `visible` channel, keeping it current as the user navigates. A block is on screen when its panel is in the **active view** and is the **active panel of its dockview group** — i.e. per-tab, so a block sitting on a background tab of the active view counts as off-screen. dock already has both signals: the active view (`client_active`) and the active panel per group (the dockview `active-panel` / `active-group` inputs, surfaced via `determine_active_views()`).

Recompute and republish on view switch, active-panel/tab change, and panel add/remove/move.

## Scope

In: blocks. Out: extensions (their servers stay eagerly constructed for now) and deferring whole-view dock construction — both later passes.

## Done when

- Switching view, or switching the active tab within a view, updates the published set.
- A block on a background tab of the active view is reported off-screen; bringing its tab forward reports it on screen.
- Adding, removing, or moving a block's panel updates the set.
- With core#203 enabled, on-screen blocks compute and off-screen ones suspend, end to end.

Depends on BristolMyersSquibb/blockr.core#203.

### LINKED ISSUE #219: Visibility gate opens at startup — every block evaluates before the front-end first reports (lazy-pull saves nothing on fresh load)
## Summary

On `204-lazy-pull` (+ blockr.dock `195-report-visible`), the visibility channel
works — core correctly logs `visibility update: N/M on screen` — but **off-screen
blocks still fully evaluate on a fresh page load**. The eval gate starts *open*,
so every block renders and pulls its result before the dock ever reports its
layout. For a one-shot dashboard load there is zero eval saving; on a large board
it's a regression (more work than the previous gating).

Repro is mechanical and shows in a 3-block app — see timeline below.

## Environment

- blockr.core `204-lazy-pull` @ `f1aa311` (contains `203-visibility-gating`)
- blockr.dock `195-report-visible` @ `0046381`
- Set `BLOCKR_LOG_LEVEL=debug` to see the gate (it logs at `log_debug`).
- Related: BristolMyersSquibb/blockr.core#203 (visibility gating), BristolMyersSquibb/blockr.core#204 (demand-driven eval).

## Root cause

The render observers are *created* suspended (`suspended = gated`), but the bridge
observer in `board_server.board()` maps an unset channel to "everything visible":

```r
# R/board-server.R (~line 85)
vis <- board_visible()                 # reactiveVal, starts NULL
if (is.null(vis)) rv$visible <- TRUE   # <-- NULL treated as "all visible"
else              rv$visible <- vis
```

So on the first flush `rv$visible == TRUE`, and `render_gate_observer()` resumes
every render observer:

```r
# R/block-server.R render_gate_observer()
do_render <- isTRUE(board$visible) || id %in% board$visible   # TRUE => resume all
```

Each resumed render reads `res()`, the lazy result reactive pulls its upstream
closure, and every block evaluates — including blocks with no on-screen
descendant. The dock's first real report (`board_visible()` -> a concrete set)
arrives later and suspends the off-screen blocks, but the eval pass has already
happened.

The underlying issue is a semantic collision on NULL:

- "no producer at all" (plain board, no dock) -> must render everything
- "producer attached but hasn't reported yet" (dock present, dockview JS still
  booting) -> should render nothing

Both are NULL; the code resolves NULL -> visible for the first case, which
silently breaks the second — i.e. every dock app on every load. Because R
constructs/flushes all block modules synchronously before the browser can report
a layout (panels must exist before they can be laid out), the open-gate window
always swallows a full eval pass. This contradicts the `block_server()` docstring,
which states "starting suspended so nothing renders before the front-end first
reports."

## Reproduction (minimal, 3 blocks)

`blockr.dock/dev/lazy-eval-min.R`: data -> a (ViewA, active), data -> b
(ViewB, hidden). Run with `BLOCKR_LOG_LEVEL=debug`, open the app, grep the log:

```
block data rendering resumed       <- all three resume immediately
block a    rendering resumed
block b    rendering resumed
evaluating block data              <- so all three evaluate...
evaluating block a
evaluating block b                 <- ...including off-screen b
visibility update: 0/3 on screen   <- dock reports for the FIRST time, AFTER
block data rendering suspended     <- gate closes too late
block a    rendering suspended
block b    rendering suspended
visibility update: 1/3 on screen
block a    rendering resumed        <- correct steady state
```

Expected: b (off-screen, no on-screen descendant) never evaluates.

## Impact on a real board (blockr.cdex, 99 blocks)

- First `rendering resumed` at log line 27 (15:58:49); first `visibility update`
  at line 1119 (15:59:26) — ~37 s and ~1090 lines later. The entire startup
  runs with the gate open.
- All 99 distinct blocks evaluate; 205 eval events total (each block 2-3x,
  because startup fires several board updates — URL restore, view reconcile — and
  each re-pulls).
- This is strictly more work than the prior gating approach (where only the
  visible blocks' closure evaluated), which is why startup feels the same or
  slightly worse.

## Suggested direction

Distinguish "producer pending" from "producer absent" so the gate can start
closed when a front-end will drive it:

- Have the producer (blockr.dock) declare itself at registration and seed
  `board_visible` to `character(0)` synchronously on the server, before any JS
  round-trip — so the initial `rv$visible` is "nothing visible" rather than TRUE,
  and the gate opens only for blocks the dockview later reports.
- Keep NULL -> visible only for the genuine no-producer case.

## Out of scope (separate issue)

Even with the gate fixed, cdex's ~47 s-to-interactive is dominated by per-session
construction of all 99 block modules + ~71 dock panels (creating block panel /
input validation / state-check churn). Lazy eval doesn't address construction —
that's the separate "defer hidden-card UI/module setup to first reveal" work.

nbenn:
## Summary

- Producer side of blockr.core#203: `board_server_callback` publishes the active view's on-screen block ids to core's `visible` write-channel so core can suspend evaluation and rendering of off-screen blocks. A block is on screen when its panel is the active (front) tab of its dockview group in the active view — per-tab, so a block on a background tab counts as off-screen.
- `report_visible_observer` recomputes on view switch (`client_active`) and on any active-view layout change — tab switch, panel add/remove/move — since dockview re-emits its serialized state on each. The channel is left untouched until the active view's dock first reports a layout, so core's no-producer default (all blocks visible) holds through startup.
- Pins `blockr.core` to the `203-visibility-gating` branch (#206) until it merges; revert the Remotes line to plain `blockr.core` then.
- The `dock app` serdes e2e snapshots block results, but per-tab gating (core#206's default) suspends the blocks it parks on background tabs, so that test now runs against a `dock-serdes` fixture with `gate_visibility = FALSE`. Gating is orthogonal to the serialization it checks and is covered by the new producer unit tests.

Fixes #195
Fixes #219

### DIFF
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -7,6 +7,8 @@
 #'
 #' @param board Reactive board state (list with `$board`).
 #' @param update Reactive update signal from blockr.core.
+#' @param visible Reactive write-channel from blockr.core, fed the set of
+#'   on-screen block ids so core can gate off-screen blocks.
 #' @param ... Extension server arguments.
 #' @param session Shiny session.
 #'
@@ -14,7 +16,8 @@
 #'   results.
 #'
 #' @noRd
-board_server_callback <- function(board, update, ..., session = get_session()) {
+board_server_callback <- function(board, update, visible, ...,
+                                  session = get_session()) {
   initial_board <- isolate(board$board)
 
   c_exts <- dock_extensions(initial_board)
@@ -58,6 +61,12 @@ board_server_callback <- function(board, update, ..., session = get_session()) {
       session
     )
   )
+
+  # Gate off-screen blocks from the first flush, before the client reports its
+  # layout (else core's all-visible default evaluates every block at startup).
+  visible(visible_block_ids(active_layout(initial_board)))
+
+  report_visible_observer(visible, client_active, docks)
 
   # Extensions receive active_dock — a reactiveValues that always mirrors
   # whichever view is currently active (swapped by update_active_dock).
@@ -159,6 +168,25 @@ switch_view_observer <- function(session, update, client_active) {
       }
     },
     ignoreInit = TRUE
+  )
+}
+
+report_visible_observer <- function(visible, client_active, docks) {
+
+  active_layout <- reactive({
+    active <- req(client_active())
+    req(docks[[active]])$layout()
+  })
+
+  observeEvent(
+    active_layout(),
+    {
+      ids <- visible_block_ids(active_layout())
+
+      if (!setequal(ids, visible())) {
+        visible(ids)
+      }
+    }
   )
 }
 
--- a/R/utils-ui.R
+++ b/R/utils-ui.R
@@ -64,6 +64,14 @@ determine_active_views <- function(layout) {
   rapply(xtr_leaf_id(root), identity, "character")
 }
 
+visible_block_ids <- function(layout) {
+
+  front_panels <- as.character(determine_active_views(layout))
+  block_panels <- front_panels[maybe_block_panel_id(front_panels)]
+
+  as_obj_id(new_block_panel_id(block_panels))
+}
+
 visible_exts <- function() {
   blockr_option("visible_extensions", "dag_extension")
 }
--- a/tests/testthat/test-board-server.R
+++ b/tests/testthat/test-board-server.R
@@ -8,7 +8,11 @@ test_that("board server", {
 
   with_mock_session(
     {
-      res <- board_server_callback(board_rv_1, update = reactiveVal())
+      res <- board_server_callback(
+        board_rv_1,
+        update = reactiveVal(),
+        visible = reactiveVal()
+      )
 
       expect_type(res, "list")
       expect_named(res, c("dock", "actions", "view_data"))
@@ -28,7 +32,11 @@ test_that("board server", {
 
   with_mock_session(
     {
-      res <- board_server_callback(board_rv_2, update = reactiveVal())
+      res <- board_server_callback(
+        board_rv_2,
+        update = reactiveVal(),
+        visible = reactiveVal()
+      )
 
       expect_type(res, "list")
       expect_named(
@@ -339,7 +347,12 @@ test_that("view_data() tracks a reported layout despite flush order (#243)", {
   withr::defer(if (!ms$isClosed()) ms$close())
 
   res <- with_mock_context(
-    ms, board_server_callback(board_rv, update = reactiveVal())
+    ms,
+    board_server_callback(
+      board_rv,
+      update = reactiveVal(),
+      visible = reactiveVal()
+    )
   )
 
   # Force the first read before reconcile runs: docks empty, so NULL.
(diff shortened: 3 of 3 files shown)


nbenn:
Testing core#204 (demand-driven lazy pull, stacked on #206) end-to-end against this producer with a complex board surfaced two things worth addressing on the dock side.

## 1. Startup evaluates every block before gating engages

`report_visible_observer` only reports once the client has round-tripped its layout:

```r
observe({
  active <- req(client_active())
  dk <- docks[[active]]
  req(dk)
  visible(onscreen_block_ids(req(dk$layout())))
})
```

Until that first client report lands, core's no-producer default ("all blocks visible") holds — which the docstring notes deliberately. With core#204's lazy eval the consequence is concrete: at the first reactive flush every block's render is resumed, so every block is pulled and evaluated; only afterwards does the client report arrive and suspend the off-screen ones. Debug logs from a 6-block board, active view showing one panel:

```
evaluating block iris_raw / iris_sub / mt_raw / iris_head / iris_head2 / mt_head   <- all 6
visibility update: 1/6 on screen     <- client reports, narrows
block iris_head rendering resumed    <- others now suspended (too late)
```

For an N-block board that is O(N) evaluation at startup no matter how few panels are on screen — exactly the cost the gating exists to avoid, and the opposite of core#203's "off-screen blocks neither evaluate nor render at startup" goal.

This is a producer-timing issue, not a core one: core gates correctly the instant a producer reports. `test-visibility-gating.R::"a producer gates evaluation and rendering on visibility"` writes `visible("b")` synchronously in the callback and asserts the off-screen blocks never evaluate — zero startup waste. So the fix is to report the active view's on-screen set eagerly, server-side, rather than waiting for the client.

Everything needed is already server-side: `board_layouts(board)` + `active_view(layouts)` give the active view's static layout, and you already call `determine_active_views(init_layout)` at `board-server.R:574`. Seeding `visible(<on-screen block ids of the active view>)` once at callback setup — before the client reports — would gate from the first flush, with the existing observer taking over for subsequent changes. (A trivial interim is `visible(character(0))` at setup: start dark and accept a one-round-trip blank; cheaper than evaluating everything, though the static-layout seed avoids the blank entirely.)

## 2. Tab-activation latency is one round-trip + eval

Good news first: the activation path has no artificial delay. The producer reads the raw dockview `_state` (`dk$layout()`), not the 250ms `sync_layouts_to_board` debounce, and dockview emits `_state` on active-tab change via `queueMicrotask` — so latency is one browser->server round-trip + block eval + render, about as tight as possible.

The thing to flag: with lazy eval, activating a tab whose upstream changed while it was hidden triggers a fresh evaluation at activation time (the deferred pull resolves then), so a heavy block can show a perceptible compute delay on first view. That is the inherent cost of laziness rather than a bug, but if it bites, keeping the most-recently-hidden tabs' renders warm — or prefetching the active tab's neighbours — would hide it. Noting it as possible future work, not a blocker.

For reference, the connectivity case checks out cleanly on the core side: a link change re-routes the pulled upstream automatically (new upstream starts evaluating, old goes quiescent), with no explicit visibility recomputation — there is a regression test for it in #211.

christophsax:
### Test this against the real cdex app before merge — minimal boards miss the gating bug

This PR is the **producer** side of the visibility-gating empty-render bug filed as
[blockr.dock#254](https://github.com/BristolMyersSquibb/blockr.dock/issues/254): during
relayout the `visible` channel can emit a **terminal `0/N on screen`** report while the
active view's panels *are* on screen, and core gating ([blockr.core#211](https://github.com/BristolMyersSquibb/blockr.core/pull/211))
treats that trailing `0` as authoritative → every view renders empty.

**The lesson worth encoding:** #254 passed *every* minimal-board test — a 2-view gated
fixture and synthetic boards of 11 views × 5 head-blocks all rendered fine. It only
reproduced on the **real cdex app**, because it's a timing race that needs real (slow)
blocks and a heavy active view (≈≥9 panels) to make the spurious `0/N` land **last**.
Minimal/synthetic boards are necessary but **not sufficient** coverage for this path.

**Asks for this PR:**

1. **Producer fix:** the `visible` channel should not emit a terminal `0/N` during
   relayout — the settled/trailing report must reflect what is actually on screen.
2. **Regression test (preferred, durable):** a headless shinytest2/chromote test that
   boots a heavy multi-panel board under `options(blockr.gate_visibility = TRUE)` and
   asserts the **active** dock renders after settle:
   ```js
   document.querySelector('.blockr-view-dock-active')
           .querySelectorAll('.dv-watermark,[class*=watermark]').length === 0
   ```
   (equivalently: the settled `visibility update` is not a terminal `0/N`). A
   self-contained fixture works — one view with ~10 real chart/table blocks reproduces
   it; no full cdex dependency needed. This is the test that would have caught #254.
3. **PR checklist:** *"smoke-tested against the real cdex app (or a heavy single view via
   `cedx_view_board("Lab")`) with `blockr.gate_visibility = TRUE` — active view has
   panels, no terminal `0/N` report."*

Repro harness + bisection + instrumented timeline are in #254.

nbenn:
One change request for the reveal path: on a view switch, the visibility report fires more than once for the same on-screen set, so core re-runs the whole reveal cascade once per duplicate report.

Measured on cedx (99 blocks): switching into a view emits `visibility update: N/99` twice with the **same** value (e.g. `2/99` then `2/99`, ~3s apart), and the reveal cascade re-evaluates every block 2–3×. That includes the linear `global_filter → ex_flat → ex_summary → ex_summary_tbl` chain, which has no multi-input node — so this is the duplicate report, not reactive fan-in waves. The apply path is clean (exactly 1× per block), so the gating itself is sound; this is specific to the reveal.

Cause is in `report_visible_observer` (`R/board-server.R`): its `observeEvent` keys on `active_layout()` — the raw dockview `$layout()` — which updates several times as panels mount during a switch. Each update re-runs `visible(visible_block_ids(active_layout()))` even when the derived block-id set is unchanged. `reactiveVal`'s `identical()` short-circuit doesn't catch it, because `visible_block_ids()` off two different layouts can differ in order/format while being set-equal.

The fix is a set-equality guard on the block-id set, in this same observer:

```r
observeEvent(active_layout(), {
  ids <- visible_block_ids(active_layout())
  if (!setequal(ids, visible())) visible(ids)
})
```

This also removes the startup `0/99 ↔ 2/99` init flap — same root cause (the report isn't coalesced) — so one guard covers both the reveal double-cascade here and the startup churn. Reporting the visible set semantically (order-independent) belongs in this observer rather than leaning on core to dedup the channel downstream. Not related to the dock#271/#265 board-update echo; that's a different channel.

nbenn:
Applied the `setequal` guard in `report_visible_observer` as suggested, folded into the producer commit (`3acf3de`). The observer now writes only when the block-id *set* changes, so the duplicate layout reports during a view switch — and the startup seed-vs-first-report pair — coalesce to a single visibility write.

Added a regression test alongside it: report one layout, then the same blocks in swapped panel order, and assert `visible()` is unchanged. It fails against the unguarded observer (the reorder slips past `reactiveVal`'s `identical()`) and passes with the guard, so it pins the coalescing contract rather than the happy path.

Unrelated to your change, FYI: refreshed against the re-rebased core#211 (`04898bb`) and `pak::lockfile_create("deps::.")` resolves clean.

Issue #193, opened 2026-06-04
Renaming block cause app crash


DivadNojnarg:
@nbenn I've experienced issues with the latest blockr.dock main branch.

```r
library(blockr.core)
library(blockr.dock)

serve(
  new_dock_board(
    extensions = new_edit_board_extension()
  ),
  "my_board"
)
```

1. Add any block from the edit board extension.
2. Rename the block title.
3. App crash.

```shell
Warning: Error in board_blocks: is_board(x) is not TRUE
  99: <Anonymous>
  98: stop
  97: stopifnot
  96: board_blocks
  87: observe
  86: <observer:observeEvent(update()$blocks$mod)>
   3: runApp
   2: print.shiny.appobj
   1: <Anonymous>
```

PR #202, opened 2026-06-04
Relabel the dock panel from the rename delta

nbenn:
## Summary

Follow-up to #200 (merged), which fixed the rename/add-panel crash by forwarding the live board into `manage_dock`. This addresses the one thing it left behind: after a rename, the block's dock panel keeps its old tab title.

- `manage_dock`'s rename observer relabels the panel, but it runs at `priority 0` while core's apply runs at `priority -Inf`. So when it reads `board_blocks(board$board)[[blk_id]]` the rename hasn't been applied yet and it sees the *old* name — `identical(old, old)` skips `set_panel_title`, and nothing else relabels a block panel, so the tab stays stale.
- Read the new name from the `blocks$mod` delta instead — the same value `apply_block_mod_delta()` applies. No board read, no dependence on apply ordering.
- Added a regression test: it fails against the unfixed observer (title not refreshed) and passes with the delta read.

Follows up #200 / #193.

### DIFF
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -655,11 +655,13 @@ manage_dock <- function(
 
         for (blk_id in names(blks)) {
 
-          if (!"block_name" %in% names(blks[[blk_id]])) {
+          delta <- blks[[blk_id]]
+
+          if (!"block_name" %in% names(delta)) {
             next
           }
 
-          new_name <- block_name(board_blocks(board$board)[[blk_id]])
+          new_name <- delta[["block_name"]]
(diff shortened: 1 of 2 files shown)


nbenn:
Blocked from the merge queue by #205: the queue-only `ci / check` e2e tests in `test-utils-serve.R` flakily fail at `AppDriver` init ("Shiny app did not become stable" — actually `address already in use`), and ejected this PR twice. This change is green on every other leg and locally, and its failing legs are unrelated tests (link-adding, multi-view nav). Will re-enqueue once #205 is resolved.

Issue #205, opened 2026-06-05
Flaky e2e tests in test-utils-serve.R repeatedly eject PRs from the merge queue


nbenn:
## Summary

The queue-only `ci / check` leg (`ubuntu-latest, devel, release`) intermittently fails at shinytest2 `AppDriver` startup, ejecting otherwise-green PRs from the merge queue. It has hit at least four PRs in the last day and forces manual re-enqueues; #202 was ejected twice in a row, which is what surfaced this.

## Symptom

Two e2e tests in `tests/testthat/test-utils-serve.R` error during `AppDriver$new(...)`:

- `edit board extension links blocks (e2e)` (L63)
- `multi-view nav renders one labelled entry per view (#189)` (L150)

```
Error ...: Shiny app did not become stable in 30000ms.
Caused by error in `app_wait_for_idle()`:
! An error occurred while waiting for Shiny to be stable
```

## It is not a real 30s timeout

The error fires ~0.1s after the wait begins, not after 30s — so this is a broken chromote↔app connection at startup, not the app failing to settle:

```
07:34:36.83 Creating new ChromoteSession
07:34:39.12 Waiting for Shiny to become idle for 200ms within 30000ms
07:34:39.16 Error while initializing AppDriver: ... did not become stable in 30000ms
createTcpServer: address already in use
Listening on http://127.0.0.1:3184
```

The `createTcpServer: address already in use` is the tell: a Shiny server from an earlier test in the same R session was still bound to the port the next `AppDriver` tried, so the new app could not come up and its chromote session never stabilised.

## Recurrence

merge_group (`ci / check`) results over the last day:

| PR | result |
|----|--------|
| #202 | failure → failure (2× consecutive) |
| #200 | success |
| #192 | success |
| #188 | failure → success |
| #187 | failure → success |

Failed run logs: https://github.com/BristolMyersSquibb/blockr.dock/actions/runs/27001795109 and https://github.com/BristolMyersSquibb/blockr.dock/actions/runs/27002135238

The other `ci / check` legs (`oldrel-1`, macOS, Windows) pass in the same runs, and the full suite passes locally — so it is timing/resource-sensitive, specific to the constrained devel runner, and independent of the code under test.

## Likely cause

[...]

PR #213, opened 2026-06-08
Pin shinytest2 to the navigate-before-bind fix

nbenn:
## Summary

- The intermittent "did not become stable" failures that eject PRs from the merge queue come from shinytest2 navigating before the app's port is bound: shiny prints "Listening on ..." before binding the socket (rstudio/shiny#4400), so on the loaded R-devel queue leg the navigation hits the browser error page and the recovery navigation races init's readiness/idle checks.
- Pins `nbenn/shinytest2@wait-for-app-serving`, which waits for the port to accept a connection (`pingr::is_up`, already a shinytest2 dependency) before navigating — fixing it at the source. Verified locally: a deterministic reproducer goes 0/3 → 3/3, with no regression (the dock e2e suite is green against the patched build).
- Supersedes the retry stopgap that previously sat on this branch. Temporary until the upstream shinytest2 PR merges, then drop the pin. `setup-idle-debug.R` is kept for now to confirm the fix on the queue.

Refs #205

```deps
BristolMyersSquibb/blockr.dag#125
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,7 +35,8 @@ Imports:
     jsonlite
 Remotes:
     BristolMyersSquibb/blockr.ui,
-    BristolMyersSquibb/blockr.core
+    BristolMyersSquibb/blockr.core,
+    nbenn/shinytest2@wait-for-app-serving
 Suggests:
     testthat (>= 3.0.0),
     DT,


tests/testthat/helpers.R
```diff
@@ -1,3 +1,54 @@
...
+
+  msg <- conditionMessage(cnd)
+
+  parent <- cnd$parent
+  while (!is.null(parent)) {
+    msg <- paste(msg, conditionMessage(parent))
+    parent <- parent$parent
+  }
+
+  grepl("did not become stable", msg, fixed = TRUE) &&
```
nbenn: Does shinytest2 not surface classed conditions?

tests/testthat/helpers.R
```diff
@@ -1,3 +1,54 @@
...
+# Construct an AppDriver, retrying only the init race above; every other
+# failure -- a real stability timeout included -- propagates. A failed init
+# leaks the Shiny subprocess it spawned, so stop it (the abort attaches the
+# half-built driver as `$app`) before the next attempt. Drop once shinytest2
+# fixes the navigate-before-load race upstream.
+new_app_driver <- function(..., max_tries = 3L) {
+
+  for (attempt in seq_len(max_tries)) {
+
+    res <- tryCatch(shinytest2::AppDriver$new(...), error = identity)
```
nbenn: Why not simply `try()`?

nbenn (commented):
Please note in #205 to clean this up again later. This is a *temporary* workaround and *not* a fix!

tests/testthat/helpers.R
nbenn: Checked — it doesn't, unfortunately. The CDP error is a bare `simpleError` whose only signal (`code: -32000` / "Inspected target navigated or closed") is interpolated into the message string — there's no `$code` field or class on it. shinytest2 then re-wraps it via `chromote_wait_for_condition` / `app_abort` as a generic `rlang_error` with no init-race-specific class. And a class wouldn't discriminate anyway: a genuine stability timeout aborts with the same generic class, so the only thing separating the race from a real timeout is the message text. So the message match is unavoidable here — I've added a comment to the helper documenting exactly that, so it no longer reads as an arbitrary choice.

tests/testthat/helpers.R
nbenn: Done — switched to `try(..., silent = TRUE)` + `attr(res, "condition")`. Cleaner, thanks.

PR #216, opened 2026-06-08
Capture the CDP navigation/context timeline on e2e init failures

nbenn:
## Summary

- Extends the temporary e2e flake instrumentation (`tests/testthat/setup-idle-debug.R`) to record the CDP navigation/context lifecycle, so the next merge-queue race shows the *sequence* that produced the failure, not just the page state at one instant.
- Both captured signatures — `Inspected target navigated or closed` and `reading 'log'` — point at an execution context being replaced, or lacking the injected tracer, mid-init. A point-in-time probe can't show the navigation/context timeline that got there; this records it.
- Adds a second trace on `app_init_browser_log()` (runs once per init, just before the navigate, with the chromote session in hand) that subscribes to `frameNavigated`, `executionContextCreated`/`Destroyed`/`Cleared`, and `loadEventFired`. The timeline prints alongside the existing exception dump on an init-check failure; it fires only on failure and does not perturb `AppDriver$new()` (verified locally — init still succeeds, full suite green).
- Independent of #213 (different file, no conflict); both are temporary and come out together once the upstream shinytest2 fix lands.

Refs #205

### DIFF
--- a/tests/testthat/setup-idle-debug.R
+++ b/tests/testthat/setup-idle-debug.R
@@ -1,15 +1,75 @@
 # Temporary diagnostic for the intermittent e2e "did not become stable"
-# failures. shinytest2 evaluates its init readiness / idle checks through
-# `chromote_eval()` and collapses any browser-side exception into one opaque
-# message, so a CI failure tells us nothing actionable. Trace that function's
-# exit: whenever one of those checks returns a JS exception, print the real
-# exception text and a snapshot of jQuery / readiness / page state at that
-# instant. Fires only on failure; no behavioural change. Remove once a failing
-# run has been captured.
+# failures. Two traces, both produce output only on failure; no behavioural
+# change.
+#
+#   1. `app_init_browser_log()` runs once per init, just before the navigate
+#      and with the chromote session in hand. Trace it to attach CDP listeners
+#      that record the navigation/context lifecycle (frameNavigated,
+#      executionContext created/destroyed/cleared, loadEventFired).
+#
+#   2. shinytest2 collapses any browser-side exception from its init readiness
+#      / idle checks into one opaque "did not become stable". Trace
+#      `chromote_eval()` on exit: when an init check returns a JS exception,
+#      print the real exception text, a snapshot of jQuery / readiness / page
+#      state, and the recorded lifecycle timeline -- the sequence a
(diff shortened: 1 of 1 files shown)


tests/testthat/setup-idle-debug.R
```diff
@@ -48,23 +108,30 @@ if (requireNamespace("shinytest2", quietly = TRUE)) {
...
+          }
           message("----------------------------------------------------")
         }
       })
     )
   }
 
   tryCatch(
-    suppressMessages(install_idle_trace()),
+    suppressMessages(install_traces()),
```
nbenn, line 124: Why do we need to `suppressMessages()` here?

tests/testthat/setup-idle-debug.R
```diff
@@ -48,23 +108,30 @@ if (requireNamespace("shinytest2", quietly = TRUE)) {
...
     error = function(e) {
-      message("[idle-debug] trace not installed: ", conditionMessage(e))
+      message("[idle-debug] traces not installed: ", conditionMessage(e))
     }
   )
 
   withr::defer(
-    suppressMessages(suppressWarnings(
+    suppressMessages(suppressWarnings({
```
nbenn, line 131: Here too? Why is this suppression needed?

nbenn (commented):
Lgtm.

tests/testthat/setup-idle-debug.R
nbenn: These only silenced the `"namespace:shinytest2"` line that `trace()` prints when it modifies a function in a namespace — harmless chatter, not worth the wrapper on throwaway diagnostic code. Removed. The instrumentation is now folded into #213 (which combines the retry + this), so I'm closing #216 in favour of that.

tests/testthat/setup-idle-debug.R
nbenn: Same here — `untrace()` prints the same line, and the `suppressWarnings` only guarded an `untrace()` of a never-installed trace (if install failed). Both removed in #213.

nbenn:
Superseded by #213 — folded this instrumentation in there alongside the retry (combining the two temporary measures into one queue attempt, per the suggestion), with the `suppressMessages` removed per the comments above. Closing in favour of #213.

PR #227, opened 2026-06-15
Pre-render the add/append block browsers; thin action handlers

DivadNojnarg:
## Summary

Pre-render the add and append block browsers and make the block action
handlers thin adapters over `blockr.ui::block_browser_server()`, which now
returns ready-to-apply `blocks` / `links` objects.

- The add and append catalogues are registry-based (board- /
  source-independent), so each is pre-rendered once into a dedicated
  sidebar (`add_block_sidebar` / `append_block_sidebar`, mounted at the
  matching action's composed namespace) and the handler just toggles it
  open — no rebuild on every open.
- `add` / `append` / `prepend` handlers apply the ready objects directly;
  removed the dock-side `build_block_from_spec()`, `valid_block_id()` and
  `valid_link_id()` (validation + port resolution now live menu-side).
- Prepend stays dynamic: its target-input picker is target/board-specific,
  and it has no UI trigger today.

Fixes the per-open append rebuild (blockr.dock#224).

## Why (profiling)

On a live deployment (54-block registry) append open was ~500 ms and
reopen was just as slow (no caching). The client render of the 50 cards was
~4 ms, so the cost was entirely the server rebuild — dominated by
instantiating every registered block to compute the linkable filter — now
moved to a one-time startup cost.

## Depends on

blockr.ui PR (ready-objects contract). `Remotes` points at
`blockr.ui@feat/block-browser-prerender-and-ready-objects`; revert to
`BristolMyersSquibb/blockr.ui` once that PR merges.

## Verification

`R CMD check` 0/0/0; `lintr` clean; `test-action-block.R` rewritten to the
new contract; verified on deployment (append now opens instantly).

### DIFF
--- a/R/action-block.R
+++ b/R/action-block.R
@@ -37,31 +37,28 @@ add_block_action <- function(trigger, board, update, ...) {
 append_block_action <- function(trigger, board, update, ...) {
   new_action(
     function(input, output, session) {
-      sidebar_id <- NS(isolate(board$board_id), "actions_sidebar")
-      # Append is trigger-specific (the source is the right-clicked
-      # block), so the browser is rendered per open. Passing board +
-      # target makes it return `list(blocks, links)` with the link's
-      # port already resolved - this handler just applies both.
+      # The append catalogue (linkable blocks) is registry-based, not
+      # source-specific, so its body is pre-rendered once in `board_ui()`
+      # (the `append_block_sidebar` mount) and opening is a pure toggle.
+      # The source (right-clicked block) is supplied here via the `target`
+      # reactive and resolved into the link at commit; the "Append from X"
+      # context goes in the sidebar title.
+      sidebar_id <- NS(isolate(board$board_id), "append_block_sidebar")
       added <- blockr.ui::block_browser_server(
         "browser",
         board = reactive(board$board),
         target = reactive(blockr.ui::append_to(trigger()))
       )
 
-      browser_ui <- function() {
-        blockr.ui::block_browser_ui(
-          session$ns("browser"), board$board,
-          blockr.ui::append_to(trigger())
-        )
-      }
+      title <- function() paste0("Append from ", trigger())
 
       observeEvent(trigger(), {
-        blockr.ui::show_sidebar(
-          sidebar_id, title = "Append new block", ui = browser_ui()
-        )
+        blockr.ui::show_sidebar(sidebar_id, title = title())
       })
 
       observeEvent(added(), {
+        # `added()` is `list(blocks, links)` with the link's port already
+        # resolved menu-side; apply both.
         res <- added()
         update(list(
           blocks = list(add = res$blocks),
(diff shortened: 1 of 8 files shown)


R/action-block.R
```diff
@@ -1,36 +1,30 @@
...
-        if (is.null(new_blk)) return()
-
-        update(list(
-          blocks = list(
-            add = as_blocks(set_names(list(new_blk), spec$id))
-          )
-        ))
+        # `added()` is a ready-to-apply `blocks` object (id-keyed, block
+        # built and named menu-side), so just add it.
+        update(list(blocks = list(add = added())))
```
DivadNojnarg, line 22: @nbenn as for stacks and links, we build the block object in blockr.ui

R/action-block.R
```diff
@@ -43,62 +37,36 @@ add_block_action <- function(trigger, board, update, ...) {
...
+        # resolved menu-side; apply both.
+        res <- added()
         update(list(
-          blocks = list(
-            add = as_blocks(set_names(list(new_blk), spec$id))
-          ),
-          links = list(
-            add = as_links(set_names(list(new_lnk), spec$link_id))
-          )
+          blocks = list(add = res$blocks),
```
DivadNojnarg, line 64: block and link built in blockr.ui

tests/testthat/test-action-block.R
```diff
@@ -364,102 +313,45 @@ test_that("prepend: NULL target_input falls back to only slot", {
...
       ))
-      expect_length(r_update(), 0L)
+
+      expect_length(r_update(), 1L)
+      expect_length(keep_calls, 1L)
+      # Add now targets its dedicated, pre-rendered sidebar and re-opens
+      # without `ui` (no re-render).
+      expect_identical(keep_calls[[1L]]$id, "my_board-add_block_sidebar")
+      expect_identical(keep_calls[[1L]]$args$title, "Add new block")
+      expect_null(keep_calls[[1L]]$args$ui)
```
nbenn, line 354: Nice — this locks down that the add handler re-opens its dedicated sidebar with `ui = NULL` (the no-re-render path). The append handler does the same thing (`append_block_sidebar`, `ui = NULL`, `action-block.R:68`) but has no equivalent assertion. Since this test is a ready template, a parallel append case would cheaply guard both halves of the pre-render toggle.

nbenn (approved):
LGTM — clean refactor. The add/append catalogues now pre-render once and the handlers thin down to applying the ready `blocks` / `list(blocks, links)` from `block_browser_server()`. I checked the parts the unit tests can't see: the pre-render↔server namespace wiring holds (same `board$board_id` contract as `actions_sidebar`), the append catalogue is genuinely source-independent (filtered by mode, source resolved at commit), and the new tests drive the real `block_browser_server` with direct payload assertions. CI is green.

The one inline note (append toggle lacks the add-toggle's assertion) is non-blocking, and the broader e2e gap for this flow is tracked in #228 rather than here.

As you noted in the description, the `blockr.ui` Remotes pin reverts once #15 lands.

tests/testthat/test-action-block.R
DivadNojnarg: Added an extra server test.

Issue #228, opened 2026-06-16
Broaden e2e (shinytest2) coverage across user-facing flows


nbenn:
blockr.dock has a small but solid e2e foundation — `test-utils-serve.R` drives real `shinytest2::AppDriver` apps (the `inst/examples/` boards plus dedicated `tests/testthat/apps/edit-link` and `apps/edit-stacks`), with DOM-fact helpers in `helpers.R` and the idle-debug setup in `setup-idle-debug.R`. We should grow it to cover the major user-facing flows, most of which today have only `testServer` unit coverage or none.

The case for e2e here is specific, not generic: dock's interactive flows live in the composition of server-rendered HTML, the dockview client, and the sidebar JS toggles — seams a `testServer` test cannot observe, since it drives inputs directly and mocks the UI. When an action handler or a layout path gets refactored, the unit tests stay green while the live wiring can silently break, and an e2e per flow is the regression guard for exactly that.

Currently covered: add-block via the edit-board extension (#191), edit-link, panel visibility on a second add (#196), edit-stacks, and multi-view nav labelling (#189).

**Scope — dock-owned behaviour only.** These tests assert dock behaviour driven through dock-owned UI: the edit-board extension (`ext-edit.R`), the block card's `⋮` menu (the dock-wired items), the view nav (`view_nav_ui` / `view-binding.js`), and the dockview client. Flows whose interactive UI lives in **blockr.ui** — the block browser (add / append / prepend block) and the settings sidebar — are blockr.ui's to test in its own suite; the revdep checks guard that blockr.ui changes don't break dock. An e2e for those in dock would mostly drive blockr.ui code, so we don't add one. On that basis #230 (append) and #235 (board options) were dropped, and #228's original navbar-add / prepend gap stays UI-less feature work rather than test work.

This is a tracking issue — one child per flow, one PR each, deliverable in priority order:

- [ ] #231 — removal flows (remove block, link, stack)
- [ ] #232 — view lifecycle (rename, remove, switch active view)
- [ ] #233 — serialization round-trip through a live app
- [ ] #234 — dock layout edits (drag / rearrange panels)
- [ ] #236 — board variants (locked-dock, single-page, views)

Conventions for these tests, matching the existing ones:

- Assert specific DOM facts — tab/panel ids, counts, label text, uniqueness — parsed with `xml2`, never UI snapshots, which are brittle, blind, and self-delete on a mid-run failure here.
[...]

Issue #196, opened 2026-06-04
Panel visibility issue


DivadNojnarg:
@nbenn I recorded a video

```r
library(blockr.core)
library(blockr.dock)

serve(
  new_dock_board(
    extensions = new_edit_board_extension()
  ),
  "my_board"
)
```

I add a block, then I want to add another one. However, somehow the block panel disappear, I only see the edit board extension panel.

https://github.com/user-attachments/assets/31ae63b6-24c6-4658-8c05-26131f48e65f

I have a related issue in blockr.dag, where the panel behavior is strange and the block ui is gone at some point (see second video).

https://github.com/user-attachments/assets/f672ddad-fb08-4e1f-8650-6409f8b4fdab

```r
library(blockr.dag)
library(blockr.core)
library(blockr.dock)

options(
  "g6R.mode" = "dev",
  "g6R.directed_graph" = TRUE,
  "g6R.layout_on_data_change" = TRUE
)

serve(
  id = "board",
  new_dock_board(
    extensions = new_dag_extension()
  )
)
```

DivadNojnarg:
On the second video, adding the head block hides the dataset panel which is not desired. Then the head block UI never appears on the corresponding panel.

Issue #191, opened 2026-06-04
Flaky `dock app` shinytest2 test: input-binding timing under load, self-deletes snapshots


nbenn:
## Summary

The `dock app` test in `tests/testthat/test-utils-serve.R` is flaky: under load it intermittently fails partway through — typically at `app$click("my_board-edit_board_extension-confirm_add")` or a following `app$wait_for_idle()` — while every other test in the suite passes. It surfaced repeatedly while landing #189 / #190, failing on roughly half of local full-suite runs but passing in isolation, at low load, and in CI.

## Why it's worth fixing beyond the flake itself

The test asserts via three `app$expect_values(export = TRUE)` snapshots. When it fails partway through, the later `expect_values` calls never run, so testthat treats their snapshots as orphaned and **deletes** `tests/testthat/_snaps/utils-serve/dock-002.json` and `dock-003.json`. A flaky failure therefore also dirties the snapshot tree, and each hit needs a manual `git checkout -- tests/testthat/_snaps` to recover. That second failure mode is self-inflicted by the snapshot approach, not just timing.

## Proposed direction

Harden the test along the DOM-assertion lines introduced in #190 (the `read_view_nav()` helper + `xml2`):

- Replace the three `expect_values` snapshots with targeted assertions on the rendered DOM (block / extension panel presence, link state) parsed via `xml2`. This removes the snapshot-deletion failure mode entirely — there are no snapshot files left to orphan.
- Give the `AppDriver` a generous explicit `timeout`, and prefer `wait_for_js` on a specific post-condition over a bare `wait_for_idle()`, which the dock's repeated busy/idle cycling makes flaky under load.

The new `multi-view nav` test in the same file is the model: DOM-fact assertions, no snapshots, explicit timeout.

nbenn:
Another occurrence while merging #187 (fixes #179), with a **distinct failure mode** worth folding into this issue.

#187's merge-queue run was ejected: the queue-only `ci / check (windows-latest, release)` job failed with `[ FAIL 1 | WARN 0 | SKIP 0 | PASS 698 ]`, the lone failure being this `dock app` test. Re-enqueue passed with no code change. So it does fail **in CI** — specifically on the `merge_group` ref, where the full `R CMD check` matrix runs (it shows "skipping" on the PR branch). [Failed run.](https://github.com/BristolMyersSquibb/blockr.dock/actions/runs/26937630580/job/79471071869)

[...]

PR #237, opened 2026-06-17
Add e2e test for appending a block via the context menu

### LINKED ISSUE #230: e2e: append block via the block context menu
Part of #228.

The block card `⋮` menu's "Append block" item (`<board>-block_<id>-append_block`) opens the actions sidebar with blockr.ui's block browser in append mode; committing a block type there adds a new block linked downstream of the source. That path — context-menu trigger → block-browser commit binding → sidebar body-swap → board update → dock render — is invisible to `testServer`. Drive it on a pre-seeded fixture: open a block's `⋮`, click "Append block", pick a block type, commit, and assert a new `block_panel-<id>` tab appears and a link from the source to the new block was created.

Append is the only wired path of #228's add/append/prepend gap — there is no navbar add-block trigger and no per-card prepend item (the `add_block_action` / `prepend_block_action` handlers are registered but nothing fires them), so those are feature work rather than test coverage. Conventions per #228.

nbenn:
## Summary

- Adds an end-to-end (shinytest2) test for appending a block via the block-card context menu — the first flow split out of the #228 e2e-coverage epic.
- Drives the seam `testServer` can't observe: a source block's `⋮` menu fires "Append block", which opens the pre-rendered append browser; clicking a catalogue card fires the browser's client-side commit binding, which the append action turns into a board update. Asserts a second `block_panel-<id>` tab mounts alongside the source, and proves the created link end to end through data flow — the appended head block computes `head(iris)` (six rows of the source's five columns), unreachable unless it received the source's output. A plain add (no link) would leave the new block input-less, so that assertion genuinely discriminates append from add.
- Adds a dedicated `apps/append-block` fixture and hoists the `block_panel_tabs()` DOM reader into `helpers.R` (was inline in the #196 test, now shared). Assertions are parsed-DOM facts via xml2 plus the board's `result` test export — no UI snapshots.
- Scope: #228's "add / append / prepend" gap is only wired for append in the current UI — there is no navbar add-block trigger and no per-card prepend item (the handlers exist but nothing fires them), so those are feature work tracked under #228, not this PR.

Fixes #230

### DIFF
--- a/tests/testthat/apps/append-block/app.R
+++ b/tests/testthat/apps/append-block/app.R
@@ -1,0 +1,13 @@
+library(blockr.core)
+library(blockr.dock)
+
+# Fixture for the append-block e2e test: a board pre-seeded with a single
+# source block, so the test drives only the append flow from that block's
+# context menu (the block-card dropdown's "Append block" item). Mirrors
+# apps/edit-link and apps/edit-stacks.
+serve(
+  new_dock_board(
+    blocks = c(a = new_dataset_block("iris"))
+  ),
+  "my_board"
+)
--- a/tests/testthat/helpers.R
+++ b/tests/testthat/helpers.R
@@ -85,6 +85,18 @@ read_view_nav <- function(app, board_id = "my_board") {
   )
 }
 
+# Block-panel tabs present in the live dock, keyed by their `block_panel-<id>`
+# id. A panel's open tab outlives its content (dockview detaches inactive
+# panel bodies), so the tab strip -- not the detached card -- is where panel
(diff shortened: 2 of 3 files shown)


nbenn:
Closing unmerged — see #230.

This append e2e primarily drives blockr.ui's block browser and sidebar; dock's append logic is already unit-tested (`test-action-block.R`) and the panel render is covered by #191/#196, so the test is effectively exercising blockr.ui's code rather than dock's. blockr.ui's own suite plus the revdep checks cover that. Dropping #230 and this PR together as part of the #228 audit.

PR #244, opened 2026-06-17
Fix dock layout save: restore reconcile priority so view_data() isn't stuck NULL

### LINKED ISSUE #243: Save/restore loses live panel rearrangement — view_data() stuck NULL, serialize falls back to default layout
## Summary

In any `dock_board` app (repro: `blockr.dm/dev/example-value-filter.R`),
rearranging panels and then **Export → Import** restores the *original*
(default) layout, not the rearranged one. The bug is on the **save** side:
the exported JSON already contains the default layout. Restore is faithful.

Regression introduced by commit **`7a36146`** *"Remove the priority=100 hack;
guard live_view_data against pending docks"* (2026-06-03, refs #164), part of
the *"Dissolve the dock manager; fold view init into reconcile"* refactor
(`f0d9244`, same day). On `main` since then.

## Root cause

`serialize_board.dock_board()` (`R/utils-serdes.R`) serializes the **live**
dock state via `view_data()`, not the board's stored layout:

```r
layout_data <- view_data()
... blockr_ser(x, ..., layouts = layout_data)
```

`blockr_ser.dock_layouts()` silently falls back to the board's stored layout
`x` when `data` isn't a `dock_layouts`:

```r
lys <- if (!missing(data) && is_dock_layouts(data)) data else x
```

`view_data()` (`live_view_data`, `R/board-server.R`) is **permanently stuck at
NULL** for the whole session, so the fallback always fires and the *default*
layout is what gets written. The same NULL also means a live drag is never
mirrored into `board_layouts(rv$board)` during the session.

Why `view_data()` is frozen at NULL:

1. On the init flush, the layout-upsync reads `live_view_data` **before**
   `reconcile_views` has populated the `docks` registry.
2. With `docks` empty, `live_view_data` hits its `return(NULL)` guard, taking a
   reactive dependency **only on `client_views()`** — it never reaches
   `dk$layout()`, so it never depends on the dock's `_state` input.
3. `reconcile_views` then populates `docks` and writes `client_views(state)`,
   but that write is `identical()` to the seed (`seed_view_state` `bare_view`
   == reconcile `bare_view`) → `reactiveVal` no-op → no invalidation.
4. `live_view_data` never re-evaluates. Frozen at NULL.

The removed `priority = 100` on the reconcile observer was exactly what forced
`docks` to be populated before the upsync first read `live_view_data`. The
commit message claims *"flush ordering no longer matters"* because the upsync
`req()`s past NULL — but it does matter: the early NULL return skips the
`_state` dependency, and the no-op `client_views()` write never re-triggers it.

## Reproduced (live, Playwright, value-filter example)

Moved `block_panel-adsl` into its own split group. Live dock = 3 groups
(`[dag][data,flt_df,flt_dm][adsl]`). Export still emitted the 1-group default
(`[dag]{panels:[data,adsl,flt_df,flt_dm]}`). Instrumentation showed
`view_data() is.null = TRUE` at serialize, and `live_view_data` returning NULL
with empty `docks`, firing exactly once (never after the move).

## Minimal fix (branch `fix/dock-layout-upsync-priority` — commit `386c53a`)

Restore `priority = 100` on the reconcile `observeEvent(board$board, …)` in
`board_server_callback` so `docks` is populated before the upsync first reads
`live_view_data`. **Verified live**: after the fix, the same rearrange exports
`[dag]{panels:[data,flt_df,flt_dm]}[adsl]` (matches the live dock) instead of
the default.

Note: although the bug lives in `board-server.R` which is identical on `main`,
this branch is cut off `feat/block-browser-prerender-and-ready-objects` (not
`main`) because the rest of the load_all'd dev tree expects that branch's
`plugin-block.R` conditions rewrite — branching off `main` reverts it and
crashes block rendering (`closure not subsettable`, plugin-block.R:532). Land
on `main` together with / after that branch, or cherry-pick once compatible.

## Proper fix (follow-up)

Make `live_view_data` re-evaluate once docks exist regardless of flush order —
e.g. a `docks`-version `reactiveVal` bumped in `create_view()` that
`live_view_data` reads — so it no longer depends on observer ordering.

## Second-order risk (separate)

Even with `view_data()` working, Export reads a **250 ms-debounced** sync; a
save fired <250 ms after the last drag can still miss it. Consider having
Export read the freshest live layout directly.

christophsax:
Fixes #243.

Restores `priority = 100` on the reconcile `observeEvent(board$board, …)` in `board_server_callback` (`R/board-server.R`), lost in `7a36146` *"Remove the priority=100 hack"*.

Without it, the layout-upsync reads `live_view_data` on the init flush **before** `reconcile_views` populates the `docks` registry. The early `return(NULL)` guard then takes a dependency only on `client_views()` (never on the dock `_state` input), and reconcile's later `client_views(state)` write is an `identical()` no-op — so `view_data()` stays NULL for the whole session. `serialize_board.dock_board()` then falls back to the board's default layout, and live rearrangements are lost on Export.

Restoring the priority makes reconcile populate `docks` before the upsync's first read, so `live_view_data` binds to `_state` and tracks live edits.

**Verified live** (Playwright, `blockr.dm/dev/example-value-filter.R`): rearrange panels → Export now captures the live layout (`[dag]{data,flt_df,flt_dm}[adsl]`) instead of the 1-group default.

Note: this branch was originally cut off `feat/block-browser-prerender-and-ready-objects`; that branch has since merged to `main` (#227), so this targets `main` directly. See #243 for the full write-up and the proposed flush-order-independent follow-up.

### DIFF
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -49,12 +49,25 @@ board_server_callback <- function(board, update, ..., session = get_session()) {
   # Reconcile the live dock session against the committed board (create /
   # destroy / restore / rename / switch views), keeping apply_board_update a
   # pure reducer. No ignoreInit: initial render is the empty-`docks` case.
+  #
+  # priority = 100 so this runs before the layout upsync on the init flush.
+  # The upsync reads live_view_data, which iterates `docks`; until reconcile
+  # has populated `docks`, live_view_data returns NULL early WITHOUT taking a
+  # dependency on any dock's `_state` input. reconcile's later client_views()
+  # write is an identical() no-op, so live_view_data would never re-evaluate
+  # and view_data() would stay NULL for the whole session -- stranding the
+  # upsync (no live layout ever mirrored to board_layouts) and the export
+  # (serialize_board then falls back to the board's default layout). Ordering
+  # dock creation first lets live_view_data's first read see the docks and
+  # bind to `_state`. Do not drop this without making live_view_data
+  # re-evaluate on dock creation by some other means.
   observeEvent(
     board$board,
     reconcile_views(
       board, update, docks, active_dock, client_active, client_views,
       committed_layouts, session
-    )
+    ),
+    priority = 100
   )
 
   # Extensions receive active_dock — a reactiveValues that always mirrors


nbenn:
Superseded by #245 (merged), which fixes #243 at the root: the dock registry is now a `reactiveValues` rather than a plain environment, so `view_data()` re-evaluates when reconcile creates a view's dock — flush-order-independent — instead of restoring the removed `priority = 100` reconcile hack. Closing in favour of that approach.

PR #253, opened 2026-06-23
WIP: Panel navigator — the "Blocks" sidebar (self-contained in blockr.dock)

christophsax:
> **WIP** — opening early for review of the approach. Verified live end-to-end; no dedicated unit tests yet (see *Follow-ups*).

## Panel Navigator — the "Blocks" sidebar

A right-side sidebar that lists **every** block on the board grouped by its
blockr.core stack (or *Ungrouped*), each row carrying a visibility **switch**
that reflects whether the block's panel is on the **active view**. It turns the
add-panel picker inside-out: instead of *"what can I add"* (the `setdiff` modal),
it shows *"here is everything you have, and where it is."*

Design spec: `_blockr.design/open/panel-navigator/` (motivation → requirements →
design → implementation).

## What it does
- **Open** via a new navbar **Blocks** button.
- Each stack is a coloured, collapsible **bar** with an **"N shown"** count;
  rows show a category-tinted icon tile, the name, and a switch.
- **The switch is the only visibility control** — drag and rename never flip it.
- **Click a visible row** → reveal its tab (`select_block_panel`).
- **Double-click a name** (block or stack) → inline rename.
- **Drag the grip** → move a block between stacks / reorder within a stack;
  drop on *Ungrouped* to clear its stack.
- **+ New stack**, search, an Ungrouped bucket, other-view tags.
- A **pinned** navigator stays live — it re-renders on a view switch and on
  external structure changes, but not on a plain toggle (keeps scroll/collapse).

## Architecture — self-contained in blockr.dock
This is a **reimplementation** of the earlier `feat/panel-navigator-stacks`
prototype, which split a "brain" into blockr.ui with blockr.dock as a thin
adapter. Here the navigator lives **entirely in blockr.dock** and needs **no
blockr.ui changes** — it builds on blockr.ui's existing sidebar primitives
(`sidebar_ui` / `show_sidebar` / `sidebar_state`) only. The CSS / JS (variant-B
"Visible blocks" design) are ported verbatim from the prototype and are fully
self-contained (no block-browser dependency).

- Visibility is a **view** concern (the dock layout, not the core board), so the
  switch routes through `show/hide_block_panel` on the active dock.
- Rename / regroup / new-stack go through the blockr.core `update()` channel as
  **deltas**, so the DAG extension and the assistant stay consistent and stack
  block order serialises.

## Files
- `R/panel-navigator.R` — observer, event handlers, model, and the variant-B renderer.
- `R/board-ui.R` — the navbar **Blocks** button + the sidebar mount.
- `R/board-server.R` — `panel_navigator_observer()` wired into `board_server_callback()`.
- `inst/assets/css/panel-navigator.css`, `inst/assets/js/panel-navigator.js` — self-contained.
- `dev/panel-navigator-demo.R` — a `load_all` harness (stack + ungrouped + two views).

## Follow-ups
- Unit tests for `nav_build_model` / `nav_reassign_payload` / `apply` handlers.
- Secondary entry point: open from a dock group's **+** (places new panels in that group).
- The agreed app-wide inline-rename widget should migrate the other rename sites.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -60,6 +60,7 @@ Collate:
     'board-plugins.R'
     'board-server.R'
     'board-ui.R'
+    'panel-navigator.R'
     'dock-board.R'
     'dock-stack.R'
     'ext-class.R'
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -181,6 +181,23 @@ live_view_data <- function(client_views, docks, client_active) {
         return(NULL)
       }
       out <- dockview_to_layout(ly)
+
+      # FIX #3: do not fold a mid-flight echo into the board. The browser's
+      # `_state` echo can fire while `apply_layout_diff` is tearing the dock
+      # down and rebuilding it, reporting a *partial* panel set. `live_panels`
+      # is the authoritative server-side membership; if the echo disagrees, it
+      # is mid-relayout -- folding it would corrupt `target` and make reconcile
+      # do another full teardown (the membership-diff loop). Hold this view
+      # (-> whole sync holds) until its echo membership catches up. Legitimate
+      # membership changes update `live_panels` via the add/remove touchpoints
+      # *first*, so they are not lost here.
+      if (!setequal(as.character(layout_panel_ids(out)),
+                    as.character(isolate(dk$live_panels())))) {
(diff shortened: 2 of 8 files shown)


christophsax:
outdated, the minidag (WIP) in blockr.outline is a better way to the same (and more)

Issue #255, opened 2026-06-24
Slow view switches on Safari (~3-4s, instant on Chrome) — output suspend/render cost on reveal


christophsax:
Spun off from #252 (the reconcile/"loads forever" loop and the dockViewR resize storm are fixed; this is the remaining, separate item).

## Symptom

With the reconcile loop and the client-side resize storm both fixed, **switching between dock views still takes ~3-4s on Safari, while it's instant on Chrome**. This includes switching *back* to a view that was already visited — so it's not a one-time cost.

## Why it's not the reconcile loop / not the server

- The server-side view switch is **~2ms**, and R is **idle between switches** (verified via `/proc/<Rpid>/stat` utime+stime deltas during the 3-4s).
- No `[loopdbg] … membership-diff -> PUSH` activity during the switch — the board is settled.

So the 3-4s is **client-side**.

## Likely cause

The revealed view's outputs resume from suspend (`suspendWhenHidden`) and **re-render their ECharts canvases** on reveal. The cedx views carry many ECharts widgets; Safari renders/initializes those canvases roughly **3× slower than Chrome**, which is where the wall-clock goes. (We do **not** yet fully understand the 3× Safari factor — keep that as an open question.)

## Fix direction (architectural — dock / core call)

- **Output suspend policy:** keep a view's outputs rendered after its first reveal (don't re-suspend on hide), so revisiting a view is instant. Trade-off: more live outputs = more memory / background render cost.
- and/or **chart render cost:** reduce per-canvas init/render cost on reveal.

This is a `blockr.dock` / `blockr.core` design decision rather than a localized bug fix.

## Repro

The full cedx board (`BristolMyersSquibb/blockr.cdex`, `run_cedx()`) on Safari — switch between e.g. Population / AE / Lab views and time the reveal. Chrome for comparison shows it's engine-bound on the client.

nbenn:
@christophsax can we close this? Or is there still something to look into here?

christophsax:
No, nothing special on Safari anymore, thanks!

PR #256, opened 2026-06-24
fix: break the dock layout-reconcile feedback loop ("loads forever" on slow clients)

### LINKED ISSUE #252: dock_board tears down all panels on load (startup race in live_view_data sync)
## Summary

On a multi-view `dock_board`, all panels render on load and are then torn down a moment later, leaving every view showing the empty "Start by adding a panel" prompt. It is a startup race, so it does not fire every time: it reproduces reliably when a data source is slow (DB / web API) and there are several panels across multiple views, and intermittently otherwise.

blockr.dock 0.1.2, dockViewR 0.3.0, blockr.core 0.1.3, R 4.4.1, macOS.

## Root cause

`board_server_callback()` wires a live-layout sync:

```r
view_data <- live_view_data(client_views, docks, client_active)
sync_layouts_to_board(view_data, update, board)   # debounced 250ms
```

`live_view_data()` reads each dock's *live* dockview layout via `dk$layout()` (`dockViewR::get_dock(proxy)`) and only skips a view when that layout is `NULL`:

```r
ly <- dk$layout()
if (is.null(ly)) return(NULL)   # guards NULL, but NOT an empty grid
out <- dockview_to_layout(ly)
...
if (any(lgl_ply(v_list, is.null))) return(NULL)
```

During startup a dockview can report a non-`NULL` but **empty** grid (it has sent its initial `_state` but `restore_layout()` has not yet run on the `initialized` event). If the 250ms debounce in `sync_layouts_to_board()` fires inside that window, `view_data()` returns a full set of *empty* layouts. `layouts_to_board_observer()` then diffs them against the board and emits `update(list(views = <mod with empty layouts>))`. `apply_views_mod()` writes those empty layouts over the board's real layouts (`board_layouts<-`), and the `observeEvent(board$board, reconcile_views(...))` handler then sees `target` layouts with zero panels and calls `apply_layout_diff()`, which removes every panel from every dock.

Traced call chain (confirmed):

```
observeEvent(board_update())
  -> apply_board_update.dock_board
    -> apply_views_mod(upd$views$mod)        # mod = {viewA: [], viewB: [], viewC: []}
      -> `board_layouts<-`                    # board layouts now empty
observeEvent(board$board)
  -> reconcile_views
    -> reconcile_view_layout(target = <empty>)
      -> apply_layout_diff                    # hides + removes all live panels
```

In a non-reproducing run, the same `apply_views_mod` receives the correct full layouts (2 / 8 / 11 panels) instead of empty ones. The only difference is timing.

## Reprex

A slow data source widens the window so the race fires (no extensions, plugins or custom blocks beyond a deliberately slow data block):

```r
library(blockr.core)
library(blockr.dock)

# stand-in for a slow DB / API source
new_slow_data_block <- function(...) {
  new_data_block(
    function(id) moduleServer(id, function(input, output, session) {
      list(expr = reactive({ Sys.sleep(3); quote(iris) }), state = list())
    }),
    function(id) htmltools::tags$div("slow data"),
    class = "slow_data_block", ...
  )
}

blocks <- c(list(data = new_slow_data_block()))
for (i in 1:9) blocks[[paste0("h", i)]] <- new_head_block()
links <- lapply(1:9, function(i) new_link("data", paste0("h", i), "data"))

board <- new_dock_board(
  blocks = do.call(c, blocks),
  links  = do.call(c, links),
  layouts = list(
    a = dock_layout("data", "h1", "h2", "h3", name = "A"),
    b = dock_layout("h4", "h5", "h6", name = "B"),
    c = dock_layout("h7", "h8", "h9", name = "C")
  )
)
serve(board)
```

Expected: each view shows its panels.
Actual (intermittently; reliably with a slow source + several panels/views): panels appear briefly, then every view collapses to "Start by adding a panel".

## Suggested fix

`live_view_data()` should treat a view as "not ready" (and suppress the sync) while its live dockview is empty but the dock is still meant to hold panels. The empty case is distinguishable from a genuinely empty view because the dock's `live_panels()` is non-empty during the race:

```r
not_ready <- vapply(seq_along(v_list), function(i) {
  dk <- docks[[names(state)[[i]]]]
  lp <- isolate(dk$live_panels())
  length(layout_panel_ids(v_list[[i]])) == 0L && length(lp) > 0L
}, logical(1))
if (any(not_ready)) return(NULL)
```

This keeps a genuinely-empty view working (its `live_panels()` is empty too) and only defers the sync until the dockview has restored. Alternatively, `layouts_to_board_observer()` could refuse to apply a `views$mod` that would reduce a non-empty board layout to empty unless the change is user-initiated.

christophsax:
Fixes #252. Prod-confirmed (Connect, exercised via Edge) — see the [analysis comment on #252](https://github.com/BristolMyersSquibb/blockr.dock/issues/252#issuecomment-4788282670).

## Problem

The full ~99-block cedx board **"loads forever"** on slow / CPU-contended clients (reliably in Safari and on production via Edge; never on local Chrome). It's a **timing race, not a browser engine** — reproduced deterministically by CPU-throttling Chromium (`Emulation.setCPUThrottlingRate`).

A mid-relayout `_state` echo from the browser gets folded into the board, which triggers a full view teardown, which emits more mid-flight echoes → repeat. Instrumenting a live Safari session showed `reconcile … membership-diff -> PUSH` firing **138× and never stopping**, all through the *membership* branch. The offending echoes are **not only empty grids** — they're **partial** panel sets captured mid-teardown (e.g. 3 of 5 panels), so an empty-grid-only guard doesn't cover the sustained loop.

## Fix

**Primary — `live_view_data()` membership guard.** Before folding a view's echoed layout into the board, require its membership to equal the dock's authoritative `live_panels()`; otherwise it's mid-relayout, so hold the whole sync until it catches up. Legitimate membership changes update `live_panels` via the add/remove touchpoints first, so nothing is lost.

**Complementary — `reconcile_view_layout()` arrangement branch.** Track a per-dock `last_applied` and only re-push on an *arrangement* mismatch when `target` actually changed since we last applied it (not merely when the browser echo lags a stable target). Doesn't fix the loop (which runs through the membership branch) but removes spurious full-restores on fast clients during normal interaction.

## Validation

- Live Safari session (server-side trace): `membership-diff -> PUSH` goes **138 → 0**, board settles.
- Deterministic repro under CPU-throttled Chromium reaches a quiet steady state instead of looping.
- **Production deploy: loads to a stable board — the "loads forever" is gone.**
- Chrome (fast client) view switches stay instant — no regression.

## Companion / follow-up

- **Companion:** cynkra/dockViewR `fix/break-resize-feedback` fixes a separate client-side resize storm (`onDidLayoutChange → resize → repeat`) that also pegged R at rest on Safari.
- **Follow-up (separate):** #255 — Safari view switches still take ~3-4s (instant on Chrome); an output suspend-policy / chart-render-cost question, not this loop.

## Notes

- A deeper rewrite of `apply_layout_diff()` (replace the full teardown with a surgical diff) is the longer-term direction discussed for a separate change.
- The diagnostic `[loopdbg]` tracing used against the live session has been stripped (final commit).

### DIFF
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -181,6 +181,23 @@ live_view_data <- function(client_views, docks, client_active) {
         return(NULL)
       }
       out <- dockview_to_layout(ly)
+
+      # FIX #3: do not fold a mid-flight echo into the board. The browser's
+      # `_state` echo can fire while `apply_layout_diff` is tearing the dock
+      # down and rebuilding it, reporting a *partial* panel set. `live_panels`
+      # is the authoritative server-side membership; if the echo disagrees, it
+      # is mid-relayout -- folding it would corrupt `target` and make reconcile
+      # do another full teardown (the membership-diff loop). Hold this view
+      # (-> whole sync holds) until its echo membership catches up. Legitimate
+      # membership changes update `live_panels` via the add/remove touchpoints
+      # *first*, so they are not lost here.
+      if (!setequal(as.character(layout_panel_ids(out)),
+                    as.character(isolate(dk$live_panels())))) {
+        message("[loopdbg ", format(Sys.time(), "%H:%M:%OS2"),
+                "] sync-guard ", v_id, " echo-membership != live_panels -> HOLD")
+        return(NULL)
+      }
+
       nm <- view_name(state[[v_id]])
       if (!is.null(nm)) {
         view_name(out) <- nm
@@ -223,6 +240,10 @@ layouts_to_board_observer <- function(view_data, update, board) {
     delta <- diff_dock_layouts(current, new_layouts)
 
     if (length(delta)) {
+      message("[loopdbg ", format(Sys.time(), "%H:%M:%OS2"), "] sync echo->board: mod=[",
+              paste(names(delta$mod), collapse = ","), "] add=",
+              length(delta$add), " rm=", length(delta$rm),
+              if (!is.null(delta$active)) paste0(" active=", delta$active) else "")
       update(list(views = delta))
     }
   })
@@ -489,6 +510,7 @@ reconcile_view_layout <- function(dock, view, target, blocks, extensions) {
 
   if (!setequal(live_ids, target_ids)) {
 
+    message("[loopdbg ", format(Sys.time(), "%H:%M:%OS2"), "] reconcile ", view, " membership-diff -> PUSH")
     apply_layout_diff(
       view = view,
       target = target,
@@ -498,6 +520,7 @@ reconcile_view_layout <- function(dock, view, target, blocks, extensions) {
     )
 
     dock$live_panels(target_ids)
+    dock$last_applied(target)
 
     return(invisible())
   }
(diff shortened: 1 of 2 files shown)


nbenn:
Holding this rather than merging — using the analysis as the spec for a source-level fix. This is about the shape of the change, not the correctness of the diagnosis (which is right; the prod verification is valuable, and we're keeping the branch as the containment). Tracked in #257 (cleanup) and cynkra/dockViewR#70 (the source fix).

The membership guard (`setequal(echoed, live_panels)`) and `last_applied` are content comparisons that recognise the dock's own echo and reject it — the same shape as the guards already on `main`: `live_panels` (#220, which itself replaced the `committed_layouts` env from #204) and the `layouts_match` / `identical` short-circuits. Each shadows one facet — commit-point, membership, now arrangement — and each new failure is an echo that slips past the facets already shadowed. This is the fourth guard of that kind.

It narrows the race, it does not remove it. Both guards are `isolate()` reads of two independently-updated reactiveVals; correctness rests on `live_panels` already reflecting committed membership at the instant an echo is folded — a timing invariant maintained by eager updates at every add / remove touchpoint, not a structural guarantee. A path that does not advance `live_panels` in time folds its echo anyway, which is how each prior guard was defeated. And the membership check does not cover arrangement: a mid-relayout echo with correct membership but a transient grid passes it and is folded — that path is left to `last_applied`, which (per your note) "doesn't fix the loop". So on a slow enough client / unlucky enough flush order the window is smaller, not closed.

Root cause is one level down: `_state` carries no provenance. `dockview.js` sets it from `onDidLayoutChange`, which fires for every layout change — a user drag and a server `restore_dock` / `fromJSON` alike — so the server cannot tell its own push coming back from real input. `apply_layout_diff` pushes via a full teardown + rebuild, so one push emits a cascade of echoes; `sync_layouts_to_board` folds one, `reconcile_views` pushes again.

The lever is already on the wire: the `restore-state` handler fires `Shiny.setInputValue(id + "_restored", true)` right after `fromJSON`, and we consume it nowhere — programmatic restores are already distinguishable at the source. Suppressing `_state` for server-driven changes (cynkra/dockViewR#70) removes the class: empty, partial, and arrangement echoes are all the server's own push returning, so one cut handles all three, and the content-comparison guards delete instead of accreting (#257). Same theme as the resize storm and cynkra/dockViewR#69.

Acceptance criteria for that work are this PR's repro and validation.

nbenn:
Superseded by #260.

That PR removes the board→dock arrangement push entirely, so the reconcile feedback loop these guards contained no longer exists — there is nothing left to guard against. #259 traces why the loop only ever came from the board→dock push, not from `board_layouts` being kept current.

Issue #223, opened 2026-06-11
Make apply_layout_diff surgical (skip the full-view rebuild on programmatic layout pushes)


nbenn:
`apply_layout_diff()` pushes a view's target layout to the live dock by tearing the whole dock down (`restore_dock`) and rebuilding it, then re-attaching every block/extension UI. Its own comment flags this as the v1 shape: "v1 unconditionally restores the target layout; surgical diff paths (noop / active-tab / reorder) are a deferred follow-up that can dispatch here without touching callers."

After #217, the common interactive paths no longer reach it — a block add, the add-panel modal, a closed tab, and an extension `show_panel()` all keep the live dock and `board_layouts` in step, so `reconcile_view_layout()` is a no-op for them. The remaining caller is a genuine board→live push: a programmatic `views$mod` (e.g. an assistant adding/removing a panel, or moving one between groups) with no corresponding live op. Those still rebuild the entire view, which visibly blinks it.

The follow-up is to make the push surgical: diff the target against the live membership/arrangement and apply only the delta — add or remove the changed panels, move/reorder within the existing dock — falling back to a full `restore_dock` only for reshapes that can't be expressed incrementally. `reconcile_view_layout()` already knows the live membership (`live_panels`) and can compute the add/remove set directly; `apply_layout_diff()` is the dispatch point.

Not a correctness issue — purely a polish item to drop the rebuild flash on programmatic layout changes.

nbenn:
Obviated by #260.

It removes `apply_layout_diff` — and `reconcile_view_layout`, its only caller — entirely. With no board→dock arrangement push there is no full-view rebuild left to make surgical. See #259.

PR #258, opened 2026-06-25
Make board_layouts authoritative, cut the layout echo loop

### LINKED ISSUE #257: Make board_layouts authoritative and remove the layout echo-suppression guards
We keep adding server-side reactiveVals that each shadow one facet of the dock layout, every one there to recognise and reject the dock's own `_state` echo coming back over a channel with no provenance: `committed_layouts` (#204) → `live_panels` (#220), the `live_view_data()` NULL / empty guards, and the membership + `last_applied` guards proposed in #256. They accrete one per facet — commit-point, then membership, then arrangement — because each new failure is an echo that slips past the facets already shadowed. #252 is the current instance.

These guards narrow the race but do not remove it. Both #256 guards are `isolate()` reads of two independently-updated reactiveVals; correctness rests on `live_panels` already reflecting committed membership at the instant an echo is folded — a timing invariant maintained by eager updates at every add / remove touchpoint, not a structural guarantee — and the arrangement facet is left to `last_applied`, which (per the PR) "doesn't fix the loop". The root cause is one level down: `dockview.js` sets `_state` from `onDidLayoutChange`, which fires for a user drag and a server `restore_dock` / `fromJSON` alike, so the server cannot tell its own push coming back from real input.

Once dockViewR suppresses `_state` for programmatic changes (cynkra/dockViewR#70), the echo no longer exists and the loop's return arrow is cut at the source. That lets us flip to a single source of truth and *remove* the guards rather than extend them:

- `board_layouts` becomes authoritative; the browser is a pure view — it renders board state and reports genuine user deltas only.
- `serialize_board.dock_board` reads `board_layouts` instead of the round-tripped `view_data()` echo. After a suppressed restore the echo no longer mirrors truth, and the server's own state is the correct source anyway.
- `reconcile_view_layout()` decides whether to re-push from server intent (what it last pushed) rather than reading and trusting the browser echo — removing the racy content comparisons.
- Delete the echo-rejection guards: the `live_view_data()` empty / partial membership guard and `last_applied`-as-echo-lag. Keep `live_panels` as the membership authority and the user → board fold; those are the genuine data path, not echo suppression.
- Move `apply_layout_diff()` toward a surgical diff (#223) so a membership change does not tear down the arrangement — also removing the cascade that amplifies the echo.

Acceptance: #252's CPU-throttle repro and the cedx board load to a stable, populated board with no feedback loop; save still captures live rearrangement; the layout-tracking guard count goes down, not up.

Depends on cynkra/dockViewR#70. Prod-verified containment in the meantime: #256.

nbenn:
## Summary

- dockViewR's `_state` echoes both the dock's own programmatic pushes (restore, add, remove) and genuine user gestures, and folding a push's echo back into the board fed the restore/reconcile loop that empties a board on a slow client (#252). This consumes the new `_state-source` provenance (cynkra/dockViewR#71): only `"client"` gestures fold into `board_layouts`, so a server push no longer echoes back in — the loop's return arrow is cut at the source.
- `board_layouts` becomes the single source of truth. `serialize_board.dock_board` reads it directly instead of the round-tripped `view_data()` echo (which could strand on a mid-flight `NULL`), and `reconcile_view_layout()` decides a re-push from the layout each view last realised (`last_applied`) plus a membership comparison, never the racy browser echo.
- The per-facet echo-rejection machinery (`live_view_data()`, `sync_layouts_to_board()`, `diff_dock_layouts()`) is removed; the layout-tracking guard count goes down, not up.
- A live membership change (a block add, the built-in `+` / `x` tab buttons) is recorded as realised rather than rebuilt, so reconcile stops fighting the live insert that just mounted the card (#191, #196). The surgical `apply_layout_diff` for programmatic membership changes stays open at #223.

Depends on cynkra/dockViewR#71, pinned in `Remotes`; drop the pin once it lands on a dockViewR release.

Fixes #257

### DIFF
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -10,8 +10,7 @@
 #' @param ... Extension server arguments.
 #' @param session Shiny session.
 #'
-#' @return List with `dock`, `actions`, `view_data`, and extension
-#'   results.
+#' @return List with `dock`, `actions`, and extension results.
 #'
 #' @noRd
 board_server_callback <- function(board, update, ..., session = get_session()) {
@@ -33,10 +32,11 @@ board_server_callback <- function(board, update, ..., session = get_session()) {
   # registry, keyed by view id; `client_active` / `client_views` mirror which
   # views the browser shows, diffed by reconcile_views() — the sole mutator.
   # Per-view panel membership lives on each dock proxy (`live_panels`), kept
-  # authoritative by the add/remove touchpoints, so reconcile compares against
-  # it rather than the lagging browser echo. `docks` is a `reactiveValues`, so
-  # live_view_data() depends on each view's entry and re-evaluates when
-  # reconcile creates it, whatever the init flush order.
+  # authoritative by the add/remove touchpoints; the arrangement each dock last
+  # realised lives on `last_applied`, so reconcile pushes from server intent
+  # rather than reading the lagging browser echo. board_layouts is the single
+  # source of truth: each dock folds only genuine user gestures (provenance
+  # `"client"`) back into it, so a server push never echoes into the board.
   docks <- reactiveValues()
   active_dock <- reactiveValues()
   client_active <- reactiveVal(NULL)
@@ -61,9 +61,6 @@ board_server_callback <- function(board, update, ..., session = get_session()) {
   # Extensions receive active_dock — a reactiveValues that always mirrors
   # whichever view is currently active (swapped by update_active_dock).
   dock <- active_dock
-  view_data <- live_view_data(client_views, docks, client_active)
-
-  sync_layouts_to_board(view_data, update, board)
 
   # Each extension can read the others' server results through the
   # `extensions` environment: one active binding per id, each resolving to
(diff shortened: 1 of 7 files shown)


R/board-server.R
```diff
@@ -563,15 +449,24 @@ manage_dock <- function(
     # on the browser's `n-panels` echo.
     proxy <- set_dock_view_output(session = session)
     live_panels <- reactiveVal(as.character(layout_panel_ids(init_layout)))
+    last_applied <- reactiveVal(init_layout)
```
nbenn, line 452: Do we really need this? We're trying with this PR to cut down on layout state tracking rv's.

R/board-server.R
```diff
@@ -563,15 +449,24 @@ manage_dock <- function(
...
     prev_active_group <- reactiveVal()
     active_group_trail <- reactiveVal()
     n_panels <- reactive(length(live_panels()))
 
+    # `layout` is the browser's live `_state` echo; `layout_source` is its
+    # companion provenance flag (`"server"` for our own pushes, `"client"`
+    # for a user gesture). Both read straight off the dockview inputs.
+    layout <- reactive(dockViewR::get_dock(proxy))
+    layout_source <- reactive(input[[dock_input("state-source")]])
```
nbenn, line 461: Do we need this? Can we not have `layout` be a reactive that only triggers on client-initiated changes?

R/board-server.R
```diff
@@ -599,6 +496,36 @@ manage_dock <- function(
...
     )
 
+    # Fold a genuine user rearrangement (drag, sash resize, tab click) into
+    # board_layouts, the single source of truth. `arrangement_fold` keeps only
+    # `"client"`-sourced echoes: a `"server"` echo is our own push coming back,
+    # and folding it would feed reconcile and loop (#252). Recording the
+    # realised layout on `last_applied` stops the reconcile this fold triggers
+    # from pushing it straight back. Debounced because a drag bursts. Membership
+    # is left alone (adds / closes round-trip server-side, reading `"server"`).
+    live_layout <- shiny::debounce(layout, millis = 250)
```
nbenn, line 506: I'm getting dizzy here, we have `live_layout`, `layout`, `live_panels`, `last_applied`, etc. The intent here was to get rid of this mess. Not to multiply it further.

nbenn:
Superseded by #260, which fixes this at the root rather than taming the echo.

#260 removes the board→dock arrangement push (`reconcile_view_layout` / `apply_layout_diff`) instead of gating the client's `_state` echo with provenance. With only the dock→board fold left there is no feedback loop to suppress, so the `_state-source` provenance, `last_applied`, and the dockViewR pin this PR introduced are all unnecessary. #259 has the full reasoning.

PR #263, opened 2026-06-25
Persist the full managed board-option set on save

nbenn:
## Summary

On an empty dock board the settings sidebar shows a "Preview rows" control, but `n_rows` is contributed by registered blocks via `blockr_app_options()` — a wider set than the board's own `board_options()`. `serialize_board.dock_board` persisted only the board's own options, so a user's change to a block-backed option was never written and reverted to the registry default (50) on reload.

- Widen the persisted set to `blockr_app_options(x)` (board + extension + block + registry options) before serializing, matching what the sidebar actually manages. On restore `combine_board_options()` is first-wins, so the saved value takes precedence over the registry default.
- Regression test for the `n_rows` save → restore round-trip, verified to fail on the prior code.

Core's `serialize_board.board` carries the same narrow/wide mismatch for non-dock boards — filing that as a follow-up.

Fixes BristolMyersSquibb/blockr.core#238

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # blockr.dock (development version)
 
+* Board options changed in the settings sidebar are no longer reset to
+  their defaults on save and reload (blockr.session#23). Serialization
+  persisted only the board's own `board_options()`, but the sidebar
+  manages the wider `blockr_app_options()` set -- options contributed by
+  blocks and the registry, such as the preview-row count -- so a changed
+  block-backed option was dropped and reverted on restore.
+  `serialize_board()` now persists the full managed set.
+
 * Live panel rearrangements are no longer lost when a board is saved
   (#243). `view_data()`, the live dock layout that serialization reads,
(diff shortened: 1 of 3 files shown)


nbenn:
Follow-up for the non-dock twin of this bug (`serialize_board.board`) plus factoring the widening into a shared core helper that both serialize methods adopt: BristolMyersSquibb/blockr.core#237. This PR stays the dock-side fix (the reprex is a dock board, and `serialize_board.dock_board` overrides core without `NextMethod()`).

nbenn:
Closing — this fixes the symptom at the wrong layer (widening inside `serialize_board.dock_board`).

The root cause is that a board never gains the options the settings sidebar manages (`blockr_app_options()`), so serialization persists only its narrow `board_options()`. Fixing that at the source — augmenting the board's options once, at `resolve_board()` in core — makes serialization correct everywhere with no dock change. Tracked in BristolMyersSquibb/blockr.core#238.

Issue #215, opened 2026-06-08
Serve the dock board closer to ready (cut post-connect assembly churn)


nbenn:
## Context

Surfaced while diagnosing the e2e merge-queue flake (#205), but independent of it. Companion to #214 (initial UI build cost): that issue is the server-side render cost, this one is the post-connect assembly shape.

Numbers below are local, R release; the CI devel runner is slower.

## What

The initial UI ships an empty container (`board-ui.R:124`, `view_container`). The board is then assembled after the socket connects, by the reconcile pass: `create_view()` (`board-server.R:325`) `insertUI`s each view's dock container and mounts the `dockViewR` widget, panels are added via proxy, `move-element` (`show-hide-block.js`) relocates block DOM, and `switch-view` (`view-binding.js`) activates — the last with a 1 s `setInterval` retry ("element may not exist yet, insertUI in same flush").

A local startup trace shows ~6 `shiny:busy`/`idle` cycles before the app settles (a typical app settles in ~1), with `empty_prompt` rendered twice.

## Direction

The post-connect assembly is by design (the reconcile pass), but the init case currently goes through the same fully-dynamic path. Consider rendering the active view's dock container (and ideally its panels) in the initial `board_ui` output so the reconcile pass only applies deltas — collapsing the cascade toward a single flush and removing the `setInterval` ordering hack in `switch-view`.

## Why

Faster, more deterministic startup; and it narrows the navigate→load window the #205 init race lands in (helpful, not a fix for #205).

PR #265, opened 2026-06-26
Expose view_data to extensions; retire redundant layout plumbing

### LINKED ISSUE #264: Expose live `view_data` to extensions; retire the redundant arrangement fold and `active_dock` mirror
## Context

#260 made a view's arrangement one-directional (dock → board): the board → dock push (`reconcile_view_layout`) is gone, so arrangement now flows only through the live-sync fold. That removal leaves several pieces of the dock-server layout plumbing redundant, and it surfaces a gap — extensions still cannot read the live, all-views layout.

Two findings frame the cleanup:

- `serialize_board.dock_board` already reads the live `view_data()` reactive, not `board_layouts(x)` (`utils-serdes.R`). Core threads `view_data` into the serializer through the callback-result → `preserve_board` plugin path, so there is no board-only serialize path.
- The only live `board_layouts(board$board)` reader in the stack is blockr.assistant (`tools-view.R`, `system-prompt.R`). blockr.ui / session / uat read it zero times, and nothing downstream reads the `dock` handle.

So after #260 the dock → board arrangement fold keeps `board_layouts` current for no remaining live reader, and `active_dock` is a hand-maintained copy of `docks[[client_active()]]`.

## Plan (bundled, forward-only)

- [ ] **Expose `view_data` to extensions** — add the live all-views layout reactive to the extension-server arg list. Additive. `view_data()` is `NULL` until every view has reported once (the all-or-nothing in `live_view_data`), so document that consumers `req()` it.
- [ ] **Retreat `active_dock` to internal-only** — it is a field-for-field denormalization of `docks[[client_active()]]` (`update_active_dock`). With `view_data` exposed, extensions no longer need its layout read; keep it as the internal block-insert / placement handle and drop it from the extension surface.
- [ ] **Rename + document `client_views`** as the nav model — it is the nav-rendered set plus display names, distinct from `docks` (built modules) only during the init flush (the static-nav seeding behind #189), and it is `live_view_data`'s name source. Full elimination (set from `docks`, names from `board`) is possible but reopens #189 and couples `live_view_data` to board churn, so it is not planned.
- [ ] **Drop the arrangement fold (`sync_layouts_to_board`)** — after the steps above it keeps `board_layouts` current for no live reader. The membership fold (`fold_live_membership` / `add_block_panels_to_view`) stays: it keeps `board_layouts` a valid skeleton and is cheap.
- [ ] **NEWS + a short extension-contract note** — live layout is `view_data()`, committed is `board`, `active_dock` is internal.

## Coordination

Stacked on #260. The one downstream break is blockr.assistant — migrate its two `board_layouts` reads to `view_data` (BristolMyersSquibb/blockr.assistant#60). ui / session / uat need no change.

### LINKED ISSUE #271: Investigate the startup board-update echo on multi-view boards (loop vs. cascade; lazy-independence)
## What

On a multi-view board, startup emits a burst of redundant board updates — none user-initiated. On blockr.cdex (12 views, 99 blocks, 30 dockview groups) a `debug`-level trace shows **~10-11 `update()` board updates** over ~25-35s during the post-connect assembly, each accompanied by a "restoring dockview layout" plus block-panel re-create/show, and correlated with the dockview cycling its active group across all 30 groups. The R process is busy the whole time.

This sits underneath the broader assembly-churn discussion in #215. This issue is narrower: pin **what drives the repeated board updates**, and settle whether it is a genuine feedback echo or the benign-but-suboptimal dynamic-assembly cascade #215 describes.

## What is established

- **The burst is not caused by lazy eval.** Reproduced on a clean eager build — `blockr.core@main` (8c87ed8) + `blockr.dock@main` (74fc12e), with neither `204-lazy-pull` nor `195-report-visible` — giving **11 board updates and 30 active-group cycles**, essentially identical to the lazy stack's ~10. So the board-update burst is a `main` behaviour; lazy does not produce it.
- On the lazy stack the same burst additionally flaps the visibility channel `0/99 ↔ 2/99` and re-evaluates the root pipeline ~3×, so lazy makes each cycle more expensive — but it does not create the cycles.

## Open questions — the "investigate further"

- **Feedback loop or bounded cascade?** Each board update is paired with a dockview re-restore. Does the re-restore re-emit state that commits the *next* update (a genuine echo), or is this just the reconcile pass settling in N steps? It does settle (bounded) in every run observed, but the round-trip shape is suspicious.
- **Which observer commits each update?** Candidates in `board-server.R`: `switch_view_observer` (`update(views = list(active = …))` on active-view changes), `layouts_to_board_observer` (`update(views = delta)` on live-layout changes), or the `live_panels` fold (`update(views = list(mod = …))`). Not yet pinned — a `trace`-level run would disambiguate (e.g. whether "committing live dock layout deltas back to the board" fires).
- **Any lazy interaction beyond cost?** The eager run rules out lazy as the *cause* of the updates, but the mechanism should be pinned to definitively rule out lazy participating in the loop itself.

## Suggested first steps

1. `trace`-level startup run on a multi-view board; classify each `update()` by originating observer and check for re-entrancy (update → reconcile → re-restore → update).
2. Ablate each candidate observer in turn (does cutting one collapse the burst to a single update?) to tell echo from cascade.
3. Check whether #215's proposed fix (render the active view in the initial `board_ui`) would eliminate the burst or merely shrink its surface to the active view.

## Relation to prior work

- #215 — the broader post-connect assembly churn and the likely *fix* (render the active view in the initial `board_ui`). The fix probably lands there; this issue is the root-cause / lazy-independence investigation feeding it.
- #260 — removed the board→dock arrangement push (one direction of an earlier loop). The board↔dock round-trip here is on current `main`, after that fix.
- #270 (closed) — an earlier teardown-loop theory (`layouts_to_board` commits transient-empty layouts), shown on the lazy stack from a since-rebased worktree. This issue brings fresh eager evidence on current `main` to check whether a residual board-update echo persists.

## Repro

`BLOCKR_LOG_LEVEL=debug`, serve a multi-view dock board (blockr.cdex is the stress case), drive a headless browser to the page, and count `starting board update` and `active group is now` over the first ~60s after the dockview reports `initialized`. Eager and lazy both exhibit it.

nbenn:
## Summary

Completes #264, stacked on #260. Exposes the live `view_data` reactive to dock extensions and retires the dock-server layout plumbing that #260 left redundant.

- **Expose `view_data` to extensions** — extension servers now receive the live all-views layout reactive (the same one `serialize_board.dock_board` reads), so an extension reads every view's current arrangement directly instead of folding through `board_layouts(board$board)`. `view_data()` is `NULL` until every view has reported its layout once, so consumers `req()` it.
- **Retire `active_dock` from the extension surface** — extensions no longer receive the `dock` handle. `active_dock` stays internal: the active-view mirror through which the board-level block insert / remove plugin places panels. The board-server return value (consumed by serialization and that plugin) is unchanged.
- **Drop the arrangement fold** — `sync_layouts_to_board` (with its `layouts_to_board_observer` / `diff_dock_layouts` helpers) folded the live arrangement back into `board_layouts` for no remaining live reader, since serialization reads `view_data()` directly. It is removed. The membership fold (`fold_live_membership`) stays — it keeps `board_layouts` a valid skeleton (view set, names, active view, panel membership).
- **Document `client_views` as the nav model** — the nav-rendered view set plus display names, and `live_view_data()`'s name source, distinct from the built-module `docks` registry only during the init flush. (Name kept; documentation added.)
- **Document the two handle bundles `board_server_callback` feeds** — the extension-server args and the spread return value, kept in step: the overlap is `view_data` + `actions` (consumed on both sides), and each difference is reasoned (`dock` returned for core's block plugin but withheld from extensions; `board` / `update` are core's own inputs; the peer env stays internal, core gets the resolved `ext_res`).
- **NEWS + extension-contract note** — the `@param server` doc for `new_dock_extension()` now documents `view_data` (live layout) and `board` (committed state); NEWS covers the contract change and the fold removal.

Downstream: blockr.assistant migrates its two `board_layouts` reads to `view_data` (BristolMyersSquibb/blockr.assistant#60). ui / session / uat need no change.

Stacked on #260 — base is `259-cut-layout-loop`; retarget to `main` once #260 lands.

Fixes #264

Closes #271

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -47,7 +47,26 @@
   the current arrangement of every view directly instead of folding it
   through `board_layouts(board$board)`. `view_data()` is `NULL` until
   every view has reported its layout once, so consumers should `req()`
-  it (#264).
+  it. The `dock` handle (the active-view `active_dock` mirror) is retired
+  from the extension surface at the same time: it is internal now, used
+  only by the board-level block insert / remove plugin (#264).
+
+* Multi-view boards no longer emit a burst of redundant board updates at
+  startup (#271). Restoring a multi-group layout makes the dockview client
+  transiently activate each group in turn, and the dock -> board arrangement
+  fold (`layouts_to_board_observer`) compared layouts through
+  `layout_to_spec`, whose `focus` field tracks the active group -- so each
+  cross-group focus tick read as a layout change and committed an
+  `update(list(views = ...))`, on the order of ten redundant updates before
+  the board settled (worst on large boards, e.g. 12 views / 99 blocks / 30
+  groups). That fold is removed: a view's live arrangement is now read on
+  demand -- on save, or by an extension via `view_data()` -- instead of
+  mirrored back into `board_layouts` on every dockview tick, so the burst is
+  gone. The removal is lossless, since `serialize_board.dock_board` already
+  reads the live layout from `view_data()` and the fold had no remaining
+  runtime reader; `board_layouts` still carries the committed view set,
+  names, active view and panel membership, kept current by the membership
+  fold (#264).
 
 * The dock no longer loops forever or tears its panels down on a slow
   client (#252). The live layout was held in two bindings that formed a
@@ -61,9 +80,8 @@
   view's arrangement is client-owned and now flows dock -> board only.
   `reconcile_views()` still owns what the board is authoritative over --
   which views exist, their names, the active view, and the initial layout
-  on load -- and the fold keeps `board_layouts` current for serialization
-  and live readers. With one direction live there is no echo to suppress
-  and no layout-tracking state to maintain.
+  on load. With one direction live there is no echo to suppress and no
+  layout-tracking state to maintain.
 
 * Live panel rearrangements are no longer lost when a board is saved
   (#243). `view_data()`, the live dock layout that serialization reads,
--- a/NEWS.md
+++ b/NEWS.md
@@ -42,6 +42,12 @@
   adjacent panel group is enough to re-emit the board, which is why #201
   surfaced this. The refresh now overlays the staged additions, edits and
   removals onto the refreshed applied state instead of clobbering them.
+* Dock extensions now receive `view_data`, the live all-views layout
+  reactive (the same one serialization reads), so an extension can read
+  the current arrangement of every view directly instead of folding it
+  through `board_layouts(board$board)`. `view_data()` is `NULL` until
+  every view has reported its layout once, so consumers should `req()`
+  it (#264).
 
 * The dock no longer loops forever or tears its panels down on a slow
   client (#252). The live layout was held in two bindings that formed a
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -68,8 +68,11 @@ board_server_callback <- function(board, update, visible, ...,
 
   report_visible_observer(visible, client_active, docks)
 
-  # Extensions receive active_dock — a reactiveValues that always mirrors
-  # whichever view is currently active (swapped by update_active_dock).
+  # Extensions receive `dock` (active_dock, a reactiveValues mirroring whichever
+  # view is currently active, swapped by update_active_dock) for the active
+  # view, plus `view_data` for the live all-views layout — the same reactive
+  # serialization reads. `view_data()` is NULL until every view has reported its
+  # layout once (the all-or-nothing in live_view_data), so consumers req() it.
   dock <- active_dock
   view_data <- live_view_data(client_views, docks, client_active)
 
@@ -98,6 +101,7 @@ board_server_callback <- function(board, update, visible, ...,
       board = board,
       update = update,
       dock = dock,
+      view_data = view_data,
       actions = triggers,
       extensions = peers
     ),
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -33,10 +33,13 @@ board_server_callback <- function(board, update, visible, ...,
   triggers <- action_triggers(actions)
 
   # Per-session dock state, closure-private. `docks` is the live manage_dock()
-  # registry, keyed by view id; `client_active` / `client_views` mirror which
-  # views the browser shows, diffed by reconcile_views() — the sole mutator.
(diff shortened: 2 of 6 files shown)


nbenn:
## This PR also resolves #271 (the startup board-update burst)

I investigated #271 — the ~10-11 redundant `update()`s during multi-view startup — and the committer is `layouts_to_board_observer`, the live-sync fold this PR removes.

Mechanism: `diff_dock_layouts` compares via `layout_to_spec`, which carries a `focus` field derived from `activeGroup` (omitted only for the default group `"1"`). So as dockview transiently activates each group while restoring a multi-group layout, every cross-group focus change reads as a layout `mod` and the fold commits an `update(list(views = delta))`. It is a bounded cascade, not a feedback echo — the return leg was already a no-op after #260 — and it is lazy-independent (lazy only amplifies each cycle's cost via the `0/99 ↔ 2/99` visibility flap and root re-eval). Verified headless: two `dock_layouts` differing only in `active_group` produce a `mod` delta and nothing else, while identical ones produce `list()`.

Removing the fold deletes that committer outright, so the burst goes away — and it is lossless: `serialize_board.dock_board` reads the live arrangement straight from `view_data()`, so the fold had no remaining runtime reader, which is exactly this PR's stated rationale. The practical shift is eager → lazy: the live layout is now read on save / on extension demand instead of folded back on every dockview tick, so transient active-group cycling no longer drives any board mutation.

One ask so the win is recorded:

- Could the NEWS entry credit this? The fold-removal bullet currently frames it as retiring plumbing with "no remaining live reader" — true, but the same removal also eliminates a real, user-visible startup board-update burst on multi-view boards (#271, the blockr.cdex stress case at 12 views / 99 blocks / 30 groups). A clause noting that captures the actual win, not just the cleanup.

I have added `Closes #271` to the description alongside `Fixes #264`. No code change requested: the removal is the fix, and being structural (no observer, no commits) it is self-verifying — a browser-level regression guard belongs with the broader assembly-churn work in #215, where the residual restore / active-group cycling still lives.

Issue #259, opened 2026-06-25
Cut the layout loop at the root: remove the board→dock arrangement push (reconcile_view_layout)


nbenn:
## Problem

The dock keeps the live layout in two bindings that together form a cycle: `sync_layouts_to_board` folds the dockview client's `_state` into `board_layouts` (dock → board, so the rv stays current for serialization and for consumers that read it), and `reconcile_view_layout` pushes `board_layouts` back to the widget via `restore_dock` (board → dock). With both live and no way to tell which side started a change, a `board_layouts` update that *originated at the dock* still triggers a re-push to the dock, whose echo folds back — the feedback loop. On a slow client a partial `_state` folds an impoverished layout into `board_layouts`, which the push then faithfully restores, tearing panels down (#252).

The board→dock push is the wrong half. The "dock renders the board" reconcile (`f0d9244`, #164) is correct for what the board is authoritative over — which views exist, their names, the active view, and the **initial** layout on load — but it was applied to the one facet the board is *not* authoritative over once running: the per-view **arrangement**, which the user drags. Pushing a client-owned facet board → dock is what forces the dock → board fold to keep them consistent, and the two together are the cycle. Every guard since — `committed_layouts` (#196/#204), `live_panels` (#220), the proposed `last_applied` (#256), the provenance approach in #257/#258 — has policed that cycle rather than removed it.

## Fix

Delete the board→dock arrangement push and let arrangement flow dock → board only:

- Remove `reconcile_view_layout()` and `apply_layout_diff()` (its only caller; the only `restore_dock` rebuild outside init).
- Keep `reconcile_views()`' structural job (create / destroy / rename / switch views) and the init-time `restore_layout()` — the board → dock directions the board legitimately owns.
- Keep the dock → board fold (`sync_layouts_to_board` + the membership fold) so `board_layouts` stays current for serialization and for live readers (e.g. `blockr.assistant` reads `board_layouts(board$board)` in `tools-view.R` / `system-prompt.R`).
- `live_panels` stays as the synchronous membership set for `n_panels` / the empty-dock prompt.

With only one direction live there is nothing to feed back on, so no echo to suppress: no `_state` provenance, no `last_applied`, no dockViewR dependency. A genuine *server-initiated* arrangement change has no callers today; if one is added it makes its own `restore_dock` at the source rather than relying on a generic reconcile.

[...]

PR #274, opened 2026-06-29
Server-authoritative dock layout: provenance-gated board→dock push

### LINKED ISSUE #273: Separate view structure from arrangement at the object level
## Motivation

After #264 (PR #265) the board's per-view layout is a mixed-freshness object. `board_layouts(board$board)` keeps each view's **membership** current (via the membership fold) but lets its **geometry** go stale, because the live arrangement now lives in the dockview client and is read on demand through `view_data()`. Serialization reads `view_data()`, `reconcile_views()` uses the board's geometry only as a restore seed at view creation, and the one remaining external geometry reader (blockr.assistant) is migrating to `view_data()` in BristolMyersSquibb/blockr.assistant#60 — so nothing reads the board's *live* geometry. Yet a caller can still do `board_layouts(board$board)[[v]]`, inspect the grid, and silently get stale positions. That footgun is the cost #264 documents but cannot remove.

The root cause is structural: a `dock_layout` is a grid tree whose leaves *are* the panel IDs (`layout_panel_ids()` is `grid_panel_ids()`), so membership and arrangement are fused into one object. The freshness boundary — server-authoritative structure vs. client-authoritative arrangement — cuts *through* the object, not between two objects, so it cannot be made honest with documentation alone. The package's own vocabulary already draws the line it does not yet store along: "A view is the conceptual page-level container; a layout is the panel arrangement inside a view."

## Target model

Split storage along the ownership boundary:

- The board stores **structure only** — a `views` collection where each `view` is `{ name, members }` (members being the ordered set of panel IDs belonging to the view, with no geometry), plus the active view id. Server-authoritative, always current, synchronously serializable.
- **Arrangement lives only in the client**, surfaced by `view_data()` (the current grid-bearing `dock_layouts`): positions, sizes, tab groups, active group. The single authoritative source for geometry.
- The serialized form **re-fuses** them: save is board structure (names, members) combined with the `view_data()` grids; the file still carries a grid per view, because restore needs it.

Readers then have two single-freshness sources — `board_views(board$board)` for structure (synchronous), `view_data()` for live arrangement (reactive) — and neither can be misread.

## Wins beyond cleanliness

- The membership fold becomes a set sync (board member-set follows the client leaf-set) instead of a grid rewrite, which is the natural home for the dock-only mutation paths (the add-panel modal, a closed tab, an extension toggle).
- The geometry-staleness bug class disappears structurally: there is no geometry in the board to go stale, and the active-group churn behind #271 never touches the board model again.
- The board stays serializable as pure structure regardless of whether the client has reported its layout yet.

## The one real design question: load-seed routing

How does the restored grid reach the dockview client without becoming a live board field (which would reinstate the stale mirror)? Recommended starting point: an explicit, immutable `restore_layout` per view, set only at deserialize and consumed once by `create_view` — write-once, read-once, so it stays honest despite living in the struct briefly. The alternative is a per-session `pending_layouts[view_id]` side-channel that `create_view` drains so the board never holds a grid at all; it is cleaner in the type but needs a non-board path from restore to dock. Start with the explicit seed and only reach for the side-channel if it proves leaky.

One boundary to settle alongside it: panel order. Tab and leaf order are arrangement (grid-owned); the board's `members` is an ordered-but-not-authoritative set, with the grid winning visually.

## Scope and sequencing

Blast radius: the board class and its `board_layouts` / `board_views` accessors, (de)serialization, `reconcile_views()`, the views-delta apply reducers, layout validation, `board_ui()`'s initial render, and the seed routing. This is a focused refactor with its own design pass — deliberately not folded into #264, which is forward-compatible with this work and a prerequisite for it.

Order:

1. #264 (PR #265) — establishes `view_data()` as the arrangement source and removes the dock to board fold.
2. BristolMyersSquibb/blockr.assistant#60 — moves the last external geometry reader off `board_layouts`. After it lands, nothing reads board geometry except the internal restore-seed path, which de-risks this refactor.
3. This issue — the structure / arrangement split.

A browser-level guard that startup produces no board-update churn is tracked separately in #215.

nbenn:
Restores the board→dock arrangement push that #259 removed, made safe by a provenance + membership fold guard (`keep_foldable`) — so the board, not the client, is the authoritative owner of dock arrangement. The consumer side of the dockViewR `_state-source` provenance work.

Closes #273

Settles the dock-arrangement direction as server-owned: it supersedes #273 (the client-owned object split) and conflicts with #265 (which retires the `layouts_to_board_observer` fold this PR relies on — only one of the two can land).

dockViewR is pinned to `cynkra/dockViewR@76-set-size`. Before merge: land that stack, drop the pin, and bump `Imports: dockViewR (>= 0.3.0)`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -36,6 +36,7 @@ Imports:
 Remotes:
     BristolMyersSquibb/blockr.ui,
     BristolMyersSquibb/blockr.core,
+    cynkra/dockViewR@76-set-size,
     nbenn/shinytest2@wait-for-app-serving
 Suggests:
     testthat (>= 3.0.0),
--- a/NEWS.md
+++ b/NEWS.md
@@ -55,6 +55,16 @@
   flows both ways again, but the cycle is broken by provenance rather than
   by dropping a direction.
 
+* The 250 ms live-sync debounce is removed (#271). It only rate-limited a
+  startup board-update burst: restoring a multi-group layout makes dockview
+  cycle its active group, re-emitting `_state` per tick, and the fold read
+  each as a rearrangement and committed a redundant board update (~10 per
+  assembly on a large board). Those echoes now carry `_state-source:
+  "server"` and are dropped by the same `keep_foldable` guard that breaks the
+  push loop, so the burst is absorbed at its source and there is nothing left
+  to bound. Focus stays live in `_state`, so a settled tab activation is
+  still folded and serialized.
+
 * Live panel rearrangements are no longer lost when a board is saved
   (#243). `view_data()`, the live dock layout that serialization reads,
   was stuck at `NULL` for the whole session, so Export fell back to the
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,21 +43,17 @@
   surfaced this. The refresh now overlays the staged additions, edits and
   removals onto the refreshed applied state instead of clobbering them.
 
-* The dock no longer loops forever or tears its panels down on a slow
-  client (#252). The live layout was held in two bindings that formed a
-  cycle: the live-sync fold pushed the dockview client's state into
-  `board_layouts` (dock -> board), and a reconcile step pushed
-  `board_layouts` back to the widget (board -> dock). A board update that
-  originated at the dock still triggered a re-push, whose echo folded back;
-  on a slow client a partial client state folded an impoverished layout
-  that the push then faithfully restored. The board -> dock arrangement
-  push (`reconcile_view_layout()` / `apply_layout_diff()`) is removed: a
-  view's arrangement is client-owned and now flows dock -> board only.
-  `reconcile_views()` still owns what the board is authoritative over --
-  which views exist, their names, the active view, and the initial layout
-  on load -- and the fold keeps `board_layouts` current for serialization
-  and live readers. With one direction live there is no echo to suppress
-  and no layout-tracking state to maintain.
+* The board is the authoritative owner of dock arrangement, and the board
+  -> dock push no longer loops (#252). dockViewR now tags every `_state`
+  echo with its origin (`_state-source`: server or client), so the board
+  -> dock push (`reconcile_view_layout()` / `apply_layout_diff()`) is
+  restored: `reconcile_views()` pushes a view's committed layout to its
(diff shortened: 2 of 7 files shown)


nbenn:
Reproduced an unbounded board↔dock echo loop under rapid view switching, on the cdex stack (core#259 + this PR merged with #289 + dockViewR@76-set-size).

**Repro:** load the board, then toggle Population↔Setup a few times *quickly* (don't wait for each to settle). Stop touching it — the active view starts flipping Setup↔Population on its own, R pegged at 100%, log growing ~unboundedly. A careful, settled Setup → Population → Setup does **not** trigger it, so it's a timing race, not the sequence itself (which is why it's intermittent).

**The self-sustaining unit** (debug trace), repeating every ~5–7s:

```
board update  →  ~10× "restoring dockview layout" (every view re-pushed)
              →  dockview re-cycles its groups ("active group is now 1,2,3,4…")
              →  next board update  →  …
```

In the captured window that was ~10 board updates against ~100 layout restores — roughly ten restores per board update — with the active-view field bouncing Setup↔Population.

**Attribution:** this is the board→dock arrangement push this PR restores — the one #260 removed to cut exactly this class of loop. Rapid active-view switching desyncs the board's active view from the client's, and the provenance gate doesn't suppress the resulting bounce, so the push and the commit chase each other. It's independent of core#259 (the paint-timing regression) and of the startup churn this PR otherwise fixes (0 board updates at rest — verified).

I attributed it from the trace (the restored push) rather than an ablation — happy to confirm by rerunning the rapid-switch repro on #289 alone (no #274) if that's useful.

nbenn:
Closing as superseded by #293, which settles the direction this PR and #273 forked on.

The settled model keeps the board authoritative for membership but makes the client the sole writer of arrangement post-init, mirrored back through settled echoes — so the normal-operation board→dock push this PR restores is exactly the edge the new model forbids (pushes happen only at dock creation and restore-as-rebuild). The `keep_foldable` guard becomes moot once the fold it protects is deleted.

Two pieces of this work survive: the dockViewR `_state-source` consumption returns as the mirror's provenance attribution (#294), and the upstream dockViewR stack this PR pinned still lands — the mirror is gated on the settled-emission end of that chain (cynkra/dockViewR#80).

PR #289, opened 2026-07-02
Build only the active view's block cards at startup

### LINKED ISSUE #272: Build only the active view's block UI at startup; defer off-screen views' cards
## Problem

`board_ui.dock_board` builds a full edit-card for **every** block — `map(block_card, board_blocks(x), …)` (`R/block-ui.R`) — across all views, into the static UI (the offcanvas mount), regardless of which view is active. On a large board this dominates first paint: on blockr.cdex (99 blocks / 12 views) the static `board_ui` render (`renderTags`) is ~20s, and it scales with the total block count, not with what is on screen.

## Why this is its own issue

- Distinct from **#214** (reduce the *cost* of the UI that is built — bslib `bs_theme()`, htmltools deps, the settings sidebar): that trims the expense of building the same tree; this is about building **fewer** cards.
- Distinct from **#215** (cut the *post-connect* assembly churn — the reconcile pass arranging already-built cards into view containers): this is the **static build** phase, before the socket connects — the off-screen cards should not be built at all.

## Direction

Render only the active view's block cards in the initial `board_ui`, and build (insert) off-screen views' cards on first open. The dynamic-insertion path already exists — `insert_block_ui()` → `insertUI()` (`R/block-ui.R`), used today when a block is added at runtime — and view-switching already shows/hides panels, so this is "don't build until shown," not new plumbing.

## Payoff

First paint pays only the active view's UI build; off-screen views cost nothing until visited. Complements #215 (less to assemble post-connect) and #214 (less to build per card), and pairs with the core-side deferral of block-**server** instantiation (separate core issue) for a fully demand-driven multi-view board.

Found while profiling cedx startup (BristolMyersSquibb/blockr.cdex#1); ~20s `renderTags` measured via Rprof of the lazy build, where eval is negligible so the figure is pure UI construction.

nbenn:
## Summary

Two coordinated changes make the multi-view board demand-driven on both axes — the UI cards and the block-server construction they feed — cutting cdex-scale startup.

### Build only the active view's cards (#272)

- `board_ui.dock_board` built an edit card for every block across every view into the static offcanvas mount, so first paint scaled with the total block count rather than what is on screen — ~20s of `renderTags` on the 99-block, 12-view blockr.cdex board.
- It now renders only the active view's cards; a per-board built-card set (carried on `active_dock`, the dock handle every card-touching path receives) tracks which cards exist, seeded to the active view at startup.
- `switch_active_view()` builds a view's not-yet-built cards on first visit (`ensure_block_ui()`) before `show_view_ui()` moves them into the now-active dock, reusing the offcanvas-insertion path (`build_block_ui()`) factored out of `insert_block_ui()`. The active-dock panel-op path (`op_add_panel()`) does the same, so a block pulled into the active view from an unvisited view is never an empty panel; `insert_block_ui()` / `remove_block_ui()` keep the set in step so a revisit never doubles a card.

### Report per-block render status on `visible` (nbenn review)

- The `visible` channel now carries a per-block status — each on-screen block tagged `required` (on screen, not yet arranged) or `rendered` (its card moved into its panel), off-screen omitted — rather than a flat id set. This gives core an explicit "the initial view is painted" signal to gate its background block-server construction on (core#259), instead of inferring readiness from result-quiescence, which mispaints at 95s on cdex.
- `report_visible_observer()` stays the sole writer, deriving `rendered` from a per-view `initialized` flag set once a view's arrangement observer has moved every card into its panel — so the channel has no two-writer race.

### Coordination

The reshape changes the `visible` wire format, so this requires core#259 (`257-render-gate`). `DESCRIPTION` temporarily pins `blockr.core@257-render-gate` plus `blockr.ui@render-gate-core-pin` (a throwaway branch repinning its own bare core remote to the same ref so `pak` resolves a single core ref instead of conflicting). Once core#259 lands on main: drop both pins and delete the throwaway `blockr.ui` branch.

Verified locally: full non-browser suite green (7203 pass, 0 fail), lint clean, multi-view app clean.

Fixes #272


nbenn:
Reshape needed, coordinated with blockr.core#259 (core consumer + motivation + measured numbers are there).

**The signal.** The reliable "the initial view is ready" point is **this observer completing for the active view** — `observeEvent(req(input[[dock_input("initialized")]]), { restore_layout(...); show_block_panel(...) }, once = TRUE)` (`board-server.R:594`). After it runs, the server has issued the whole arrangement: the layout to the dockView and every card moved out of the offcanvas into its panel. core#259 currently guesses readiness from result-quiescence and gets it wrong (95s paint on cdex); the dock should report it explicitly instead.

**The reshape:** `visible` becomes a **multi-level per-block status** — `hidden` | `required` | `rendered` (provisional names) — instead of a flat id set. Piggyback on `visible`, not a sibling channel (rationale in core#259). Core derives on-screen (`required ∪ rendered`) and ready (all `rendered`) from it.

**Dock side:**
- `report_visible_observer` (`board-server.R:174`): when a block goes on-screen, write it at `required` — this is today's `visible()` write, just tagged "on screen, not yet arranged."
- At the **tail of the `input$initialized` observer** (`board-server.R:594`), after the `show_block_panel` loop, **upgrade the arranged block ids to `rendered`**.
- Thread the status writer into the per-view module: `create_view` (`board-server.R:363`) already receives `update`; give it the `visible`/status writer so the arrangement observer can reach it. `board_server_callback(board, update, visible, …)` (`:19`) keeps receiving `visible` — now the status writer.

**Bonus:** because the dock arranges a panel regardless of whether its block's data is ready, this drops the input-waiting-block hang the result-quiescence approach needed a fallback for — the signal is "the panel is up," not "the data settled."

nbenn:
Done — dock side implemented against core#259 (`257-render-gate`) and pushed.

`visible` now carries the per-block status: `report_visible_observer` writes each on-screen block `required`, upgrading to `rendered` once its view is arranged. One deviation from the letter of the spec — rather than thread the status writer into `manage_dock` and mark `rendered` from the arrangement observer, I kept `report_visible_observer` as the *sole* writer and had it derive `rendered` from a per-view `initialized` flag the arrangement observer sets at its tail. Same signal (that observer completing for the active view), but the channel never has two writers racing on it and nothing new is threaded into the per-view module. Trivial to switch to the explicit writer-thread if you'd rather.

The #272 interaction falls out for free: off-screen views' arrangement observers run no-op moves (their cards are deferred), so only the active view's blocks flip to `rendered`; a lazily-visited view flips on first visit, when its cards are built and shown. And since the gate is one-shot (`gate$destroy()` after the initial paint), switches after that don't need to re-report `rendered` — `required` is enough for the eval-gate, which keys on membership.

Verified locally against `257-render-gate` (installed the branch): full non-browser suite green (865), and the multi-view app runs clean — active view's cards built, off-screen deferred, a switch builds them, no server errors (the stray `errors_block` warning is gone now that core defers the off-screen servers).

The reshape changes the `visible` wire format, so the PR pins `blockr.core@257-render-gate` (plus a throwaway `blockr.ui@render-gate-core-pin` repinning the same ref); `pak::lockfile_create` resolves clean. I'll drop both once core#259 lands on main.

nbenn:
Profiled the reshape now that it's stacked on the #300 loop-fix (this branch's tip `9e2009f`, sitting on `e28e04e`), against `core@257-render-gate` + `dockViewR@76-set-size`, on cdex. Two results: the loop-fix lands cleanly, but it unmasks a separate startup-starvation bug that leaves the visible view painting dead-last.

**The churn is gone.** Stacking on #300 drops the startup churn from 30 active-group cycles / 14 board updates (this branch off `main`) to **1 / 0**. That half of the problem is solved.

**But net visible paint got worse, not better** — ~145s here vs ~78s for the un-stacked branch, against a ~28s target (background construction held off entirely). The un-stacked churn was accidentally re-establishing the visible set sooner; with it gone, the underlying starvation is exposed.

| metric | un-stacked (off main) | stacked on #300 | target (bg held) |
| --- | --- | --- | --- |
| active-group cycles | 30 | 1 | — |
| board updates | 14 | 0 | — |
| net visible paint | ~78s | ~145s | ~28s |

**Mechanism** (server timeline, session start `09:21:28`):

1. `:38` — dock reports Setup's 2 blocks visible (`required`); they evaluate (`data :40`, `cdisc :53`).
2. `:55` — the arrangement restores **all 12 views' layouts** (11× "restoring dockview layout"), which leaves the dockView active group on an off-screen view → **visibility drops to 0/99** → `data`/`cdisc` **suspend**.
3. `:55` — the reshape gate sees the visible set reported `rendered` and **releases background construction**, which then hogs the single R thread `:55`→`:23:46` (59 blocks).
4. The client active-group events from step 2 — the ones that would settle visibility back to 2/99 — are **starved behind that construction**. They aren't processed until it finishes: the "active group is now 2,3,4,1…" burst lands at `:23:50`, and only then does visibility recover to 2/99 and the visible blocks render. Net ~145s.

**Root cause:** the gate releases before the visible view is *client-confirmed* stable. Two coupled defects feed it — (a) startup initializes and restores *all 12 views'* dockViews, not just the active one, so the active group is left on the wrong view; and (b) the `rendered` mark fires at the tail of the server-side `show_block_panel` loop, before the client's active-group round-trip settles. Either alone would be survivable; together, the background construction wins the race for the thread and starves the correction for the whole ~2min it runs.

**Fix:** the clean root-cause fix is to defer off-screen views' dockView entirely — only initialize and restore the active view's dock at startup, mirroring this PR's card deferral — so no off-screen dockView is ever touched to cycle the active group and the 0/99 transient never happens. If some off-screen restore must stay eager, re-selecting the active view at the tail of the multi-view restore is a narrower fallback that at least settles visibility to 2/99 before the background can starve it. core#259's gate itself is correct — it releases exactly when the visible set is reported `rendered`; the fix is to not report it `rendered` until the visible view has actually settled.

Not a stale-#300 artifact: I checked the three commits this branch is behind #300's current tip (`3e26140`) — the relevant one, "Bridge the arrangement mirror," touches sash-drag/mirror-commit semantics, not the visibility/active-group path. Re-syncing is clean (board-server.R auto-merges; only DESCRIPTION and a test conflict) but wouldn't change this result.

Issue #294, opened 2026-07-06
Route membership through the update lifecycle and arrangement through a settled-echo mirror


nbenn:
Part of #293 (design and enforcement table live there). Depends on #273 (the record split) and on cynkra/dockViewR#80 (settled `_state` emissions — pin the branch during development, bump `Imports:` on release).

## Scope

Replace the fold with the mirror, and narrow the lifecycle to membership:

- One canonical projection function (echo → stored arrangement): relative sizes normalized and quantized, defaults elided, volatile ids dropped. Idempotent; the arrangement slot setter rejects values that are not its fixed point.
- Provenance on arrangement values: `authored` (constructor, deserialization) vs `echo` (projected from `_state`). The mirror observer asserts `echo`; the creation push asserts `authored`.
- The mirror observer becomes the sole arrangement writer: its write capability is a closure handed to exactly one observer, and writes commit the settled echo verbatim in canonical form, `identical()`-guarded. No membership projection at write — the slots are related by total semantics (see #293): a member without an arrangement entry is an un-landed intent, an arrangement entry without membership is an inert ghost, and neither is acted on outside a boundary (no live reconciler on the gap).
- The update lifecycle becomes membership-only: `views$mod` carries panel-id sets, `validate_board_update.dock_board` rejects geometry payloads with a pointer to the mirror, and `merge_views_mod` / `rm_blocks.dock_board` collapse to set operations.
- Validation relaxes to match the semantics: `validate_dock_arrangements()` drops its not-a-subset rejection (one-sided states are legal transients — a block removal legally commits a ghost); board validation checks `members ⊆ blocks` (referential) plus structural grid validity, and containment pruning moves to compose/boundary hygiene.
- Deletions: `sync_layouts_to_board` and its debounce, the `diff_dock_layouts` spec comparison, `append_panel_to_layout`, and the placement synthesis in `add_block_panels_to_view`.

## Acceptance

A sash drag or tab switch produces at most one board commit and zero pushes; a membership change produces one lifecycle commit and one mirror write after the client places the panel; every committed board state is a defined row of the semantics table, and `members ⊆ blocks` holds on every commit (checked in `validate_board.dock_board`).

Issue #293, opened 2026-07-06
Layout ownership rework: single-writer model with a board-stored arrangement mirror


nbenn:
## Problem

The dock's layout state has two writers: the update lifecycle folds client echoes into `board_layouts` (synthesizing placement it doesn't own via `append_panel_to_layout`), while the client owns the actual arrangement and re-echoes after every change. Every remaining bug in this area is a symptom of that dual ownership: reconciliation loops, lost updates inside the debounce window, dead server-side geometry writes (`views$mod` arrangements never reach a live dock), and the mixed-freshness footgun #273 documents. The repo currently holds both directions as competing drafts — #274 (server-authoritative, provenance-gated board→dock push) versus #265 + #273 (client-owned arrangement, object split) — and only one can land. This issue settles the direction and tracks the rework; #274 is superseded (see below).

## Settled architecture

Two board slots, one writer each, both last-commit fresh:

- **Membership** — per view: an ordered member panel-id set plus the display name; on the collection: view ids and the active view. Written only by the update lifecycle (server ops and client intents routed through `update()`), as set operations — never tree surgery.
- **Arrangement** — per view: the canonical spec (nesting, tab groups, active tabs, relative sizes, focus). Written only by the settled client echo, verbatim, as a one-directional mirror. The server stores and serializes it but never interprets or synthesizes it.

The two slots are related by **total semantics** rather than a cross-field invariant: every (members, arrangement) pair is meaningful, so no committed or serialized state can be invalid. Membership decides *which* panels appear; the grid (né arrangement) decides only *how*. At every read boundary `view_grid()` resolves placement: the grid's geometry applies to the members it places, a member the grid omits is given a default spot (the grid never withholds), a ghost (grid panel outside membership) is dropped, and a view with no grid at all renders its members via the default grid. The one rule that keeps this loop-free: nothing acts on the members-minus-arrangement gap in a live session — the add-flow's own in-flight op resolves it, and only dock creation/restore prunes and normalizes. One deliberate identification (settled in #298): a default-shaped arrangement — an even split of its own panels, in order — canonicalizes to `NULL` for authored and echo values alike, so "default-shaped" and "never expressed" are the same stored state; the drop-unlanded-members restore rule therefore does not bind for default-shaped views (a mid-flight export there restores the pending member default-placed rather than dropped), in exchange for zero commits on the first echo of default views and real headless `NULL` persistence.

[...]

PR #297, opened 2026-07-06
Split view structure from arrangement into two board slots

### LINKED ISSUE #273: Separate view structure from arrangement at the object level

nbenn:
## Summary

- A view's `dock_layout` fused panel membership with grid geometry, so `board_layouts()` was a mixed-freshness object — the membership fold kept the panel set current while geometry went stale. That footgun is structural: the freshness boundary cuts *through* the fused object, so it can't be closed with documentation (the cost #264 documents but can't remove).
- Store the two along the ownership boundary instead: a `dock_views` collection of `dock_view` structure objects (ordered panel-id set, name, id, active view), read with `board_views()`, and a separate `NULL`-valid `dock_arrangements` geometry slot, read with `board_arrangements()`. `validate_board.dock_board` enforces `arrangement ⊆ membership`, and a board is valid with no arrangement at all.
- Scope is the record split only. `new_dock_board(layouts = ...)` still takes fused `dock_layout()`s and decomposes them at construction (public DSL unchanged), and `board_layouts()` / `active_layout()` stay as composed bridges so the fold, reducers and reconcile keep working untouched. Rewiring the flows onto the two slots — the settled-echo mirror, deleting the placement synthesis — is #294.
- Serialization carries both fields as separate wire entries; geometry is still read live from `view_data()` for freshness, and a pre-split file (a fused `layouts` field) still deserializes through a fallback.

Stacked on #265. While the base is `264-extension-view-data`, CI (gated on `branches: main`) and the `Fixes` closing link stay dormant; both activate on retarget to `main` after #265 lands.

Fixes #273


nbenn:
Heads-up from the semantics settled in #293 (total pair semantics: shown = members ∩ arrangement, one-sided states are defined transients rather than errors): the not-a-subset rejection in `validate_dock_arrangements()` is correct for this PR's bridge world, where decompose-on-write keeps the two slots exactly equal — but it must not outlive it. Under #294's flows a block removal legally commits a ghost (arrangement entry without membership) for one round trip, so the rejection there relaxes to prune-at-compose hygiene, with hard validation shrinking to `members ⊆ blocks` plus structural grid validity.

No change needed here; noting it so the containment gate isn't tightened further or copied into new call sites in the meantime.

Issue #296, opened 2026-07-06
Loop-safety enforcement: commit budget, model-based interleaving tests, e2e sentinels


nbenn:
Part of #293. Depends on #294 and #295. Converts the single-writer principles from convention into things that fail loudly, so the loop class stays dead through future refactors.

## Scope

- Dev-mode commit budget: count board commits per client event using the provenance tags; exceeding a small bound aborts with the attributed commit chain. A loop becomes a stack trace naming its two writers instead of a spinning app.
- Model-based interleaving tests: pair the pure reducers with an abstract client model (tracks panels, a tree, sizes; applies ops; emits settled echoes; places panels arbitrarily — the properties are placement-agnostic, so no dockview logic is duplicated, and arbitrary placement is a stronger test than faithful placement). Generate seeded random sequences over {server op, client gesture, echo delivery with flush lag, restore} and assert after quiescence: convergence (board arrangement equals the canonical client state), fixed point (no commit cascade), no ghost rendering (a removed panel never renders again), every interleaving landing in a defined row of the semantics table, and commit-count bounds (one gesture, at most one commit).
- Thin e2e sentinels per the house testing doctrine: expose the commit counter as a dev-mode output, run the canonical gesture set (sash drag, panel move, tab switch, block add/remove, restore), assert the expected counts and a zero-commit reading after quiescence via parsed-DOM facts.

## Acceptance

The property suite runs in CI on every PR with seeded generators (failures replayable by seed); the e2e sentinel fails on any regression that adds a second commit to a single gesture.

PR #298, opened 2026-07-06
Route membership through the lifecycle, grid geometry through a mirror

### LINKED ISSUE #294: Route membership through the update lifecycle and arrangement through a settled-echo mirror

nbenn:
## Summary

- Membership travels the update lifecycle as panel-id **sets** (`views$mod`), and `validate_board_update.dock_board` rejects a layout smuggled into `mod` with a pointer to the mirror. `rm_blocks.dock_board`, `merge_views_mod`, and the live-membership fold collapse to set algebra; `add_block_panels_to_view` is a set-add, not placement synthesis.
- Grid geometry travels a **settled-echo mirror** — the sole writer of a view's stored grid (the `board_grids()` slot, per #297's rename), a closure handed to one observer. It reads dockViewR's `_state-source` to ignore the echoes its own restores provoke, `identical()`-guards on the canonical value, and commits the settled echo **verbatim in canonical form**.
- The two slots relate by **total semantics**, not containment: a member with no grid entry is an un-landed intent, a grid entry with no membership an inert ghost. Both are legal on a committed board and reconciled only at the compose / restore boundary — ghosts pruned, an expressed grid normalised to the intersection, an unexpressed one (`NULL`) shown via a default. No live writer touches the gap.
- One canonical projection (`project_grid`): sizes quantised to a coarse grid (above resize jitter, below user intent), volatile ids / default tabs / load-default focus dropped via a wire-spec round-trip, a plain default elided to `NULL`. Idempotent; `validate_dock_grids` enforces every stored grid as its fixed point carrying `authored` / `echo` provenance.
- Validation relaxes to match: the not-a-subset rejection is gone (a block removal legally commits a ghost); `validate_board.dock_board` checks `members ⊆ blocks` (referential) plus structural grid validity instead. `append_panel_to_layout` is deleted; `serialize_board.dock_board` reads both slots straight from the last committed board.

A sash drag or tab switch is at most one board commit and zero pushes; a membership change is one lifecycle commit plus one mirror write once the client places the panel; every committed board is a defined row of the semantics table with `members ⊆ blocks`.

**Decoupled from the dockViewR stack** (tracked in #301): a 250 ms debounce bridges the mirror's `_state` feed as a server-side stand-in for gesture-settled emission, so the dependency stays a bare `cynkra/dockViewR`. The `_state-source` read is NULL-safe — the skip self-disables against a dockViewR that does not emit the companion input. Rebased onto `main`, adopting #297's `arrangement` → `grid` rename.

### `views$mod` contract

| before | after |
| --- | --- |
| `list(mod = list(v = dock_layout(...)))` — a full grid | `list(mod = list(v = c("block_panel-a", ...)))` — a membership set |

Fixes #294


nbenn (commented):
Reviewed against the semantics settled in #293 (total pair semantics, verbatim mirror). The implementation is faithful: verbatim echo commits with the `identical()` guard on the projected form, provenance stamped at every authored site and re-stamped on deserialization, `apply_views_arrange` asserting `echo` and re-projecting as defense in depth, the not-subset gate replaced by `members ⊆ blocks` plus the canonical-fixed-point check, `append_panel_to_layout` deleted, and `serialize_board` reading the stored slots. The wire-vs-slot `NULL` handling (commit the unelided canonical because the update channel is NULL-lossy, elide in apply) and the 0.5% quantization grid with largest-remainder correction both match the design intent.

One change requested, two notes for the record.

**Change — the docstring on `observe_arrangement_echo` describes the superseded design.** It says the observer "projects the client layout onto the view's current membership" and that "a panel removed since the echo cannot resurface". That is the pre-amendment behaviour this PR correctly does not implement: the code stores the echo verbatim, and ghosts are inert under total semantics, pruned only at the compose/restore boundary. A comment asserting the dead design over the live code invites a future "fix" back toward it — please rewrite those two sentences to describe the verbatim/total-semantics behaviour.

**Note (tracker) — the default-elision defines "default-shaped" as "never expressed".** A plain default is elided to `NULL` by `project_arrangement()` for authored and echo values alike. Good trade: first echoes of default views commit nothing, and headless `NULL` persistence becomes real. But it means the drop-unlanded-members restore rule never binds for default-shaped views — a mid-flight export there restores the pending member default-placed rather than dropped. Deliberate and coherent; worth one sentence in #293's semantics paragraph so it reads as a decision rather than a surprise.

**Note (#296) — the `_state-source == "server"` skip deserves a round-trip stability property.** The skip is only load-bearing when dockview normalizes a pushed layout to something canonically different, in which case the stored slot runs one gesture behind client truth until the next user gesture. The harness should assert that the authored canonical form equals the canonicalization of its own dockview echo; if that property holds the skip is pure safety, and if it ever fails we want a red test rather than silent drift.

Merge gates as already known: the `cynkra/dockViewR@76-set-size` pin drops once the upstream stack lands and releases (with the `Imports:` bump), and the base retargets to `main` after #297.

nbenn:
To merge the stack without waiting on the dockViewR chain (its timing is outside our control), this PR needs one amendment commit. The rest of the stack is verified upstream-independent: the stack introduces no new `dockViewR::` calls versus `main`, and the `_state-source` read is NULL-safe.

**1. Drop the `cynkra/dockViewR@76-set-size` pin from `Remotes:`** — back to the bare `cynkra/dockViewR`; `Imports: dockViewR (>= 0.2.1)` stays, since no new R-side API is consumed.

**2. Add a 250 ms debounce on the arrangement mirror's source** (the `dock$layout()` feed into `observe_arrangement_echo`), commented as an explicit bridge: a server-side stand-in for gesture-settled `_state` emission, to be removed in #301 once dockViewR ships the settled chain. Without it, current dockViewR emits unsettled per-frame states during a sash drag and the mirror would commit per frame — an interactive regression versus main's pre-rework cadence.

**3. Why the debounce is safe here when it was part of the disease before**: the old debounce raced a second writer and fed a spec diff; this mirror is the sole writer, commits verbatim canonical values, and diffs nothing — the only cost is 250 ms of bounded staleness on a slot whose readers are all boundaries. It reproduces main's exact commit cadence while the loop machinery stays deleted.

**4. No other changes needed.** The `_state-source` skip self-disables on current dockViewR (absent input reads `NULL`), which yields one benign, convergent commit per view at creation and restore — expected, and the seeded round-trip probe asserts the same stored-equals-projected equality either way. The e2e sentinel's bounded-plus-quiescence assertions hold under the bridge (`wait_for_idle` outlasts the window).

**5. After force-pushing, rebase the children immediately** (`296-loop-safety`, then `295-restore-rebuild`), per the stack cadence — this is the third time the base moves under them.

Removal of the bridge, the `Imports:` bump, and the optional sentinel tightening are tracked in #301.

PR #299, opened 2026-07-06
Add loop-safety enforcement: commit budget, model tests, e2e sentinel

### LINKED ISSUE #296: Loop-safety enforcement: commit budget, model-based interleaving tests, e2e sentinels

nbenn:
## Summary

Enforces the dock's layout single-writer principles — one writer per slot, so the reconciliation-loop class stays dead through future refactors — where enforcement is cheap and precise, and **without leaving anything in the running app**.

- **Model interleaving tests** (`test-model-interleaving.R`): the primary detector. The pure reducers (`apply_board_update` for membership, the settled-echo mirror, `view_grid` placement) are driven against an abstract client that renders a panel set and echoes a `dock_grid`, placing panels arbitrarily. Seeded generators walk the `{server op, gesture, echo-with-flush-lag, restore}` alphabet and assert — in test code, with no production instrumentation — board validity, no ghost rendering, membership authority (a view shows exactly its members), and at most one commit per gesture; at quiescence, convergence to what the client shows plus a re-echo fixed point. Failures replay by seed.
- **Thin e2e sentinel** (`test-commit-sentinel.R`): guards the one seam the model can't reach — the real dockView round-trip. It reads two **test-only server exports** via blockr.core's `blockr_test_exports` hook (`R/utils-serve.R`): a commit count (blockr.core's own `board$last_update$seq` tally) and a round-trip stability flag. These are registered through `shiny::exportTestValues()`, gated by Shiny's test mode, so a normal running app carries no trace of them — no output, no DOM, no option. `board-server.R` and `board-ui.R` are untouched. A settled gesture ticks the count a bounded amount and quiescence adds none; a loop either never settles (idle wait times out) or keeps the count climbing.

Division of labour: the exact "at most one commit per gesture" bound is proven deterministically by the model suite; the e2e is the seam-level spike check no unit test can observe.

#296's original scope had a runtime commit *budget* (abort-on-bound). It was cut after review: a dev-mode breaker is off in production (protects nothing where a loop would hang) and unnecessary in a test (a loop should hard-fail, not recover). What's wanted is spike *detection* — the model suite does it in test code, the e2e does it at the seam by reusing core's existing tally, and the observation rides the idiomatic `blockr_test_exports` channel rather than any production hook.

Targets `main` (#298 merged; the harness is ported to the rewritten `dock_grid` representation, and the two bugs the earlier harness had surfaced are subsumed by that rewrite). The e2e sentinel drives real dockView and only runs against Chrome in CI.

Fixes #296


nbenn (commented):
Reviewed against #296's mandate: do the teeth bite the failure modes they were designed for. Overall strong — and the best evidence is that the harness earned its keep before merging: the interleaving suite caught a real bug (an empty-membership view under a ghost-only arrangement was a legal committed board that could not serialize — `grid_to_spec` aborted on the pruned-to-`NULL` root), fixed consumer-side with the `grid_map_leaves` contract deliberately preserved and a proven fail-without/pass-with regression test.

Findings, in order — the first is mechanical and should come before the rest.

**1. Rebase onto the current base tip first — stack skew currently poisons the diff.** The base branch gained 75925e4 ("Describe the mirror's verbatim behaviour in its docstring") after `296-loop-safety` was cut, so this PR's diff reads as if it re-introduces the superseded "projects onto current membership" docstring on `observe_arrangement_echo`. The PR's own commits don't touch that hunk; a rebase clears it without conflict. Until then, a diff-reader will conclude this PR reverts the #298 fix.

**2. The abort message names the wrong unit.** The error in `record_commit` says "budget of N per client event exceeded" and the comment on `instrument_commits` says "the bound is per client event", but the mechanism is per flush — the file-top docstring has it right. One event approximates one flush cascade for the synchronous loop class, but a diagnostic that misnames its unit costs trust exactly when someone is debugging a loop. Two-line wording fix.

**3. Document the runtime tooth's scope on the option.** Catching only intra-flush synchronous spins is the right call — the round-trip loop class is dead by design (no down-path in normal operation) and the CI quiescence sentinel covers that seam — but the `dock_commit_budget` doc should say so in a sentence, so nobody assumes it guards echo loops.

**4. The round-trip stability property is still only abstract, and this PR owns the sentinel, so it lands here.** The model proves projection idempotence (a re-echo commits nothing), but the property that justifies the `_state-source == "server"` skip concerns real dockview: the authored canonical form must equal the canonicalization of dockview's echo of it. Concrete ask: expose `identical(strip_provenance(stored), strip_provenance(project_arrangement(<live layout>)))` as a test output in the commit-probe app and assert it post-load, pre-gesture. One detail: the fixture must author a non-default layout (explicit sizes), otherwise both sides elide to `NULL` and the assertion is vacuously true.

Everything else checked out: the wrapper installs before anything captures `update`, so every dock writer is attributed through one chokepoint at zero production overhead; the probe is suspend-exempt and correctly namespaced; the `force(update)` self-recursion trap is caught and commented; the model suite (40 seeds × 30 steps; per-step totality, containment and commit bound; quiescence convergence plus re-echo fixed point) matches the #296 spec; and the sentinel drives a real add through the edit extension and asserts DOM facts.

nbenn:
Thanks — all four addressed, force-pushed.

**1. Rebased.** The branch now sits on the current base tip (75925e4), so the diff no longer reads as touching the `observe_arrangement_echo` docstring — that hunk belongs to the base now, and the rebase was conflict-free as you predicted.

**2. Fixed.** The abort message and the `instrument_commits` comment both say "per flush" now, matching the mechanism and the file-top docstring.

**3. Documented.** The `dock_commit_budget` doc now states the scope: the runtime tooth is intra-flush only, the cross-flush echo loop class is dead by design (no board → dock push in normal operation), and the e2e quiescence sentinel is that class's guard — so it isn't read as guarding echo loops.

**4. Round-trip stability lands here.** Added `arrangements_stable(stored, live)` — for every view, `identical(strip_provenance(stored[[v]]), strip_provenance(project_arrangement(live[[v]])))` — exposed as a `roundtrip_stable` probe output and asserted post-load, pre-gesture. The comparison is unit-tested deterministically (a non-default authored layout stays a fixed point; a size-drifted one does not), and a new `commit-probe-seeded` fixture authors an explicit-sizes layout so neither side elides to `NULL` and the assertion isn't vacuous. The e2e reads the real dockview echo, so CI is where the property justifying the `_state-source == "server"` skip is actually exercised — the model only proved projection idempotence.

Issue #233, opened 2026-06-17
e2e: serialization round-trip through a live app


nbenn:
Part of #228.

Serialization is only unit-tested today (`test-utils-serdes.R`, pure functions). The dock-owned value an e2e adds is the live round-trip: build a board in a running app, serialize, restore, and assert the dock-owned state — `dock_layout` / `dock_layouts` / `dock_extensions`, the active view, and the producer-version routing — rebuilds and re-renders identically, alongside blocks/links/stacks. This exercises the dock serde + reconcile path through the running server, which the pure-function tests don't.

Conventions per #228.

PR #300, opened 2026-07-06
Test the restore e2e's grid render and the visibility seed

### LINKED ISSUE #295: Restore as dock rebuild; boundary reads for creation, visibility seed, and headless boards
Part of #293. Depends on #294 (the mirror). Answers the load-seed routing question #273 left open: the seed is the `authored` arrangement, and it reaches the client through exactly one door — dock creation.

## Scope

- Dock creation pushes the `authored` arrangement once (`restore_layout` on `initialized`, as today). Placement resolves at read via `view_grid()`, membership-authoritative: the grid's geometry applies to the members it places, members the grid omits get a default spot (an in-flight add at export time restores visible, never silently dropped), ghosts (grid panels outside membership) are pruned, and a grid-less view renders its members via the default grid.
- Restore becomes an explicit rebuild event: the serdes import path tears down and recreates all docks from the restored board instead of relying on the view-set diff in `reconcile_views`. This fixes the latent bug that re-importing a board whose view ids collide with the live ones silently never re-applies its layouts (the diff sees no view-set change).
- The initial `visible()` seed reads the authored arrangement at birth (its open tabs), falling back to all membership panels when the arrangement is `NULL`.
- Headless round-trip: `new_dock_board(layouts = ...)` followed by `blockr_ser()` / `blockr_deser()` with no session preserves the authored arrangement, and a membership-only board round-trips with `arrangement = NULL`.

## Acceptance

Re-importing the currently exported file into a running session reproduces the exported arrangement; a never-rendered board round-trips byte-identically; after any restore, the first settled echo populates the mirror without triggering a push.

nbenn:
## Summary

The last sub-issue of #293. Since #298 merged, the grid rework unit-tests most of #295 in the base itself — headless round-trip, ghost pruning, membership-authoritative placement, and the mirror. This narrows #295 to the two seams the base does not reach:

- **Restore round-trip (#233 e2e):** the existing test asserts nav and blocks only. This also asserts the grid — the restored `analysis` view keeps its authored a/b tab group with `b` fronted (headless), the re-export after the reload reproduces the `grids`/`views` payload byte-for-byte, and the dockview client renders `b` as the front tab (`active_block_panel_tabs()`, the client leg a server-side read cannot see).
- **Visibility seed:** the active view's open tabs — the front panel of a tab group, every panel of separate leaves.

Tests only.

Fixes #295

### DIFF
--- a/tests/testthat/helpers.R
+++ b/tests/testthat/helpers.R
@@ -237,6 +237,30 @@ block_panel_tabs <- function(app, board_id = "my_board") {
   sort(sub(".*-tab-(block_panel-.+)$", "\\1", xml2::xml_attr(nodes, "id")))
 }
 
+# The block panels whose tab is the *front* (active) tab of its dockview group,
+# scoped to one `view`'s dock (every view's dock is in the DOM, so an unscoped
+# read would mix in another view's front tabs). The client-rendered geometry a
+# server-side read cannot see: dockview marks the active tab by toggling
+# `dv-active-tab` on the `.dv-tab` wrapper of the custom tab element (id
+# `...-tab-block_panel-<id>`), so the front tab is that wrapper's descendant tab
+# id. A restored group shows exactly one front tab; a collapse to separate
+# leaves surfaces every panel as its own. The view's dock sits in the stable,
+# server-rendered `#<board>-view_handle-<view>` (not the client-toggled
+# `.blockr-view-dock-active`, which can lag).
+active_block_panel_tabs <- function(app, view, board_id = "my_board") {
+  html <- xml2::read_html(
+    app$get_html(paste0("#", board_id, "-view_handle-", view))
+  )
+  xpath <- paste0(
+    "//*[contains(@id, '-tab-block_panel-')]",
+    "[ancestor-or-self::*[contains(concat(' ', normalize-space(@class), ' '), ",
+    "' dv-active-tab ')]]"
+  )
+  nodes <- xml2::xml_find_all(html, xpath)
+  ids <- sub(".*-tab-(block_panel-.+)$", "\\1", xml2::xml_attr(nodes, "id"))
+  sort(unique(ids))
+}
+
 # Shared helpers for the edit-board extension e2e tests (links, stacks). The
 # extension namespaces its inputs under `my_board-edit_board_extension-`. The
 # extension panel stays active in those tests (pre-seeded fixtures, no block
--- a/tests/testthat/test-board-server.R
+++ b/tests/testthat/test-board-server.R
@@ -552,6 +552,42 @@ test_that("board_server_callback seeds visibility before the client reports", {
   })
 })
 
+test_that("the visibility seed reads the active view's open tabs", {
+
+  # An expressed tab group hides its back tab: only the front panel of each
+  # group seeds visibility (the active view's placed grid at birth).
+  board_rv <- board_args(
+    blocks = c(
+      a = new_dataset_block(), b = new_head_block(), d = new_head_block()
+    ),
+    grids = list(v = dock_grid(panels("a", "b", active = "b"), "d"))
+  )
+
+  with_mock_session({
+    visible <- reactiveVal()
+    board_server_callback(board_rv, update = reactiveVal(), visible = visible)
+
+    expect_setequal(isolate(visible()), c("b", "d"))
+  })
+})
+
+test_that("the visibility seed spans separate leaves", {
+
+  # Separate single-panel leaves each front their own tab, so every member is
+  # on-screen (none is a hidden back tab).
+  board_rv <- board_args(
+    blocks = c(a = new_dataset_block(), b = new_head_block()),
+    grids = list(page = dock_grid("a", "b"))
+  )
+
+  with_mock_session({
+    visible <- reactiveVal()
+    board_server_callback(board_rv, update = reactiveVal(), visible = visible)
+
+    expect_setequal(isolate(visible()), c("a", "b"))
+  })
+})
+
 test_that("view nav renders one labelled item per view (#189)", {
 
   board <- new_dock_board(
--- a/tests/testthat/test-utils-serve.R
+++ b/tests/testthat/test-utils-serve.R
@@ -306,6 +306,11 @@ test_that("a board survives the live Export/Import round-trip (#233)", {
     c("ext_panel-edit_board_extension", "block_panel-a", "block_panel-b")
   )
 
+  # Grid geometry survives deserialization, not just membership: the analysis
+  # view restores its authored a/b tab group with b fronted (a is the hidden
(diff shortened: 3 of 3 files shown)


nbenn (commented):
Reviewed against #295's amended scope. The PR's central claim — that the capstone needed almost no code because #297/#298 already made every boundary read correct — checks out, and that is the best possible verdict on the rework: the own commit (9aba476) is one dead-code removal plus ~314 lines of tests that lock the boundary behaviour in place. Both load-bearing claims verified independently: restore-as-rebuild rides core's staged reload (`board_refresh` → `loader$stage()` → `session$reload()` in blockr.core's `utils-serve.R`, so a colliding-id re-import structurally cannot strand a layout — there is no in-session diff to no-op), and the removed `extra` list in `restore_board.dock_board` was genuinely dead (`as_dock_board.dock_board` returns its input verbatim and ignores `...`), with the `meta` branch correctly kept against the reload-loop trap.

Two findings — do the first before the second, since it moves the diff under review.

**1. Rebase onto the current base tip.** The base `296-loop-safety` was rewritten underneath this branch: it was rebased (picking up 75925e4) and its loop-safety commit amended in response to review — including the new `commit-probe-seeded` app. This PR sits on the pre-rebase tip, so its branch diff falsely shows it deleting that review response. The own commit does not conflict with those files; the rebase should be clean. Cadence suggestion for the stack, since this is the second consecutive occurrence: branch from the freshly-fetched base tip at cut time, and rebase children immediately after any review-response force-push — otherwise every review cycle poisons the next PR's diff.

**2. The client leg of the restore e2e is unasserted — the one substantive gap.** The new byte comparison in the #233 test is valuable (the wire is a fixed point through a real import → reload → re-export cycle), but its comment claims it "proves the stage / reload / rebuild pipeline delivered the docked arrangement", and it cannot: since #298 the export reads the stored slots, and the creation echo is skipped via `_state-source == "server"`, so a client-side `restore_layout` failure would leave both the byte check and `read_dock_state()` green — the latter (helpers.R) reads only the nav and block-handle ids, no geometry. Concrete ask: assert one DOM fact about the imported arrangement — the fixture authors a tab group with a non-first active tab and custom sizes, so the active tab in the rendered tab strip (or a panel width ratio) closes the loop cheaply, in the house DOM-facts style. Failing that, soften the comment to "proves persistence through the reload".

Everything else checked out: the visibility-seed test pair pins the intended asymmetry (a constructor `default_layout` is a tab group — expressed, seeding visible = active tabs only — while a genuinely-`NULL` view composes to separate leaves and seeds all members); a never-rendered board round-trips byte-identically; the `NULL`-survives-active-switch test empirically retires the earlier caveat that bridge deletion was a prerequisite for `NULL` persistence; and the mirror test covers the restore provenance flow (server echo skipped, first client echo transitioning `authored` to `echo`).

nbenn:
Both addressed — the rebase went first, so this now reflects the correct diff.

**1. Rebased onto the current base tip.** Replayed my commit onto `296-loop-safety`'s new tip (`1af14be`) with `--onto`, so the branch diff no longer shows deleting `75925e4` / the commit-probe-seeded response — it's back to my own files only (+365/-9), MERGEABLE/CLEAN. Point taken on the cadence: I'll cut children from the freshly-fetched base tip and rebase them immediately after any review-response force-push, so a review cycle stops poisoning the next PR's diff.

**2. Client leg now asserted.** You're right — the byte comparison reads the stored slots and the creation echo is server-sourced (`_state-source == "server"`, skipped), so a client-side `restore_layout` failure stayed green. Added a rendered DOM fact: after the import/reload, `active_block_panel_tabs()` reads the tab strip and asserts the analysis a/b group restores with `block_panel-b` as the front tab — anchored on dockview's `dv-active-tab` (`.dv-tab` wrapper of the `...-tab-block_panel-<id>` element). A collapse to separate leaves surfaces both panels as front tabs, and active-defaults-to-first surfaces `a`; both change the set and fail. I also scoped the byte-comparison comment to "survives the stage/reload cycle" rather than overclaiming the client leg.

Honest caveat: chromote doesn't paint the dockview client here, so I couldn't drive the real render locally. I validated the helper's xpath/regex against a synthetic DOM matching dockview's structure (both the wrapper-child and same-element layouts) — it returns `block_panel-b` for a correct render and provably flips on each failure mode; CI exercises the actual render.

nbenn:
Reworked onto the reworked base and force-pushed (`39f90a8`, MERGEABLE/CLEAN).

The stack was rebased again overnight and the mirror layer was substantially reworked underneath this PR: `arrangement` → `grid` throughout (`board_grids`, `observe_grid_echo`, `canonicalize_grid`, wire slot `grids`), provenance and the `_state-source` skip dropped, `all.equal(tolerance = grid_size_tol())` replacing quantise + `identical`, and `restore_board.dock_board` dropped to inherit core's. Given the scope I reset onto the new base tip and reauthored my net-new tests against the grid model rather than force a conflict-heavy rebase.

Net effect: the PR is now **tests only**. My one-line `restore_board.dock_board` cleanup is subsumed (the base removed the whole method), and my grid-mirror test is redundant with the base's `test-grid-mirror.R` — both dropped. What remains, translated to the grid model: the headless byte-identical / membership-only-`NULL` / `NULL`-survives-active-switch round-trips, the three restore-normalization-at-load cases, the visibility-seed open-tabs-vs-all-members pair, reconcile-seeds-the-authored-grid, and the #233 e2e client-leg (re-export byte check on `grids` + the rendered front-tab assertion via `active_block_panel_tabs()`).

Full non-e2e suite 7189 pass / 0 fail on the new base; lint 0. The e2e leg stays CI-only.

nbenn:
Reauthored onto the settled base (`0150a21`, MERGEABLE — no conflict; BLOCKED only by the dormant required checks while stacked).

The #298 grid rework has merged and settled the model: `arrangement` → `grid`, the elision and the compose bridge (`board_layouts`) dropped, and placement is now membership-authoritative — `view_grid` / `place_members` default-places a member the grid omits rather than dropping it (the reverse of #295's original "restores as if never added" text). That rework carried its own comprehensive unit tests (round-trip, ghost pruning, placement, mirror), which subsume most of my earlier suite.

So I reset onto the settled tip and kept only the genuinely net-new coverage: the #233 e2e grid + client-render leg, and the visibility-seed open-tabs asymmetry. Retired: the `restore_board.dock_board` cleanup (method dropped upstream), the elision / NULL-persistence tests, and the "un-landed member dropped" normalization — the latter two assert behavior the settled model removed or reversed. Net +96 lines across 3 test files, down from ~300.

Heads-up: #295's issue text still says an un-landed member is "dropped … available in the picker"; the merged model default-places it instead — worth a one-line issue edit to match, at your convenience.

Full non-e2e suite 7108 pass / 0 fail; lint 0. The e2e leg stays CI-only.

PR #303, opened 2026-07-07
Drop bridge debounce, re-add _state-source skip; adopt move_panel

### LINKED ISSUE #301: Remove the settled-echo bridge debounce once dockViewR ships gesture-settled state
Part of #293 — the follow-up to the merge-now decoupling of the layout-ownership stack (#265 / #297 / #298 / #299 / #300) from the cynkra/dockViewR release chain.

The stack merges against current dockViewR: #298 gains a 250 ms debounce on the arrangement mirror's source as an explicit bridge — a server-side stand-in for gesture-settled `_state` emission — because today a sash drag emits unsettled per-frame states and the mirror would otherwise commit per frame. The bridge is loop-safe by construction (sole-writer verbatim LWW commits, no diffing; the 250 ms window is bounded staleness on a boundary-read slot) and reproduces the pre-rework cadence. Loop-freedom does not depend on upstream; only cadence does.

Trigger: a dockViewR release carrying the settled-emission chain (cynkra/dockViewR#79 → #83 → #71 → #80).

## Scope

- Remove the bridge debounce from the mirror source wiring and the comment marking it.
- Bump `Imports: dockViewR (>= <that release>)` so the settled cadence is a declared requirement.
- Clean the leftover comment referencing the debounced state echo in `utils-dock.R`.
- Optionally tighten the e2e commit sentinel from bounded-per-gesture to exactly one commit per settled gesture, now that the cadence is guaranteed.
- No change to the `_state-source` skip: it self-enables when the release ships the companion input, eliminating the one convergence commit per view at creation and restore; the seeded round-trip probe already asserts the equality that keeps the skip safe.

## Acceptance

Searching `R/` for debounce returns nothing; the commit sentinel is green (tightened, if adopted); the mirror unit tests pass unchanged.

### LINKED ISSUE #320: Add the `resize` panel-op verb once the set_size proxy lands
#318 (PR #319) establishes panel operations as `views$mod` verbs — `add` / `rm` / `move` / `select` — but leaves `resize` out. `resize` needs the `set_size` dockView proxy command to deliver, which is unmerged upstream (cynkra/dockViewR#78) and pinned only by the still-open #303. Rather than ship a verb that validates but silently no-ops, #319 rejects `resize` as an unknown verb, so an early caller fails loud instead of quietly getting nothing.

Adding it back is localized, mirroring the other pure-payload verbs (`move` / `select` write nothing to the board and exist only as delivery + mirror capture):

- **Validate** (`validate_view_mod`): add `resize` to `view_mod_verbs()`; each entry keys a member panel to a ratio in `(0, 1)`.
- **Apply**: nothing — geometry is client-owned and captured by the settled-echo grid mirror, exactly as for `move` / `select`.
- **Deliver** (`R/panel-ops.R`): `set_size` on the panel's group, addressing the group through a member panel (the `referencePanel` idiom the grammar already uses for placement), idempotent against the settled ratio.

Grammar shape (per the #318 design):

```r
update(list(views = list(mod = list(
  <view-id> = list(resize = list(<panel-id> = <ratio>))
))))
```

Blocked on the `set_size` proxy (cynkra/dockViewR#78, pinned by #303). blockr.assistant#64's `resize_panel` tool lands on the same floor, so the two can be picked up together.

### LINKED ISSUE #327: Grid mirror commits dockviewR's intermediate restore frames, transiently corrupting the committed grid
`observe_grid_echo()` (the settled-echo grid mirror) commits any client layout that differs from the stored grid, debouncing `dock$layout()` by 250ms. It cannot distinguish a genuine user reorder from dockviewR *still applying* a restored layout.

dockviewR applies a restored layout by streaming intermediate `_state` frames (the reason the mirror debounces at all — see the settled-`_state` bridge, #301/#303). During a restore one of those frames transiently reports a tab group as separate leaves before the back tabs collapse. If that frame sits still longer than the 250ms debounce, the mirror commits the collapsed layout to the board's authoritative committed grid, then re-commits the grouped layout once the client settles.

So for a window during load/restore the committed grid is genuinely wrong. A user who exports or saves in that window persists a flattened grid — a tab group split into separate leaves. Small window, but a real data-fidelity bug, not just a test artifact.

## How it surfaced

The core#274 two-channel visibility adoption (#289) added per-flush work (two per-block reactiveVal channels, a `priority = Inf` validate observer, a needed-derivation observer) that lengthened the restore transient past the 250ms debounce. Old core restored fast enough to stay under it, so this stayed latent. It reproduced as a flaky failure of the `#233` serdes round-trip e2e: the first run committed a collapsed Analysis a/b group (both the exported-grid artifact and the DOM), and a re-run passed. #289 mitigated the *flake* test-side (wait for dockviewR to settle before reading the exported/rendered layout), but the underlying mirror behaviour is unchanged.

## Evidence

Server-side the authored grid is correct (`visible_block_ids(active_view_grid)` returns just the fronted tab). The committed grid is the only thing that goes wrong, and the mirror's `commit_grid` is its sole writer after init — so the mirror committed a client-reported intermediate frame.

## Fix direction

Gate the mirror so it does not commit while a view's restore is in progress: suppress `commit_grid` until the arrange/restore observer has settled, so dockviewR's intermediate frames during a restore are not mistaken for user edits. This is the same restore-vs-mirror boundary the settled-`_state` work already navigates.

nbenn:
## Summary

Now that #306 and dockViewR#86 (released as dockViewR 0.4.0) have landed, this rebases onto `main` and builds on the released chain: the gesture-settled `_state` emission, the first-class `move_panel`, and the `set_size` proxy.

### Settled `_state`: drop the bridge, gate the mirror on a restore latch

- `board-server.R` — `observe_grid_echo()` reads `dock$layout()` directly (the 250 ms debounce bridge is gone) and commits every settled echo that differs from what is stored, **except** the one the initial restore provokes. `manage_dock` raises a one-shot `restoring` latch around its single `restore_layout()` call; the mirror consumes it on the next echo and skips exactly that one. A restore replays the composed layout we already hold (dockViewR coalesces it into one settled echo through its buffered `onDidLayoutChange`), so committing it would overwrite the stored grid with a pruned copy or, mid-load, persist a transient frame — the bug #327 describes.
- **Not** a blanket `_state-source == "server"` skip. dockViewR raises its server flag around *every* proxy op — `add-panel`, `move-panel`, `set-size`, `restore-state`, … — so a skip keyed on the tag also drops the settled echo of a programmatic move / resize / add, and their geometry never persists (the `resize` verb below would be inert on save). The latch is owned by blockr.dock, which knows the one op — restore — whose echo is a redundant replay; every other server op's echo is new truth and commits. (A finer-grained provenance tag in dockViewR that distinguishes restore from other ops would let the mirror lean on the tag again; that is a separate upstream change.)
- `DESCRIPTION` — bump the floor to `dockViewR (>= 0.4.0)` for `move_panel` + `set_size`. dockViewR#86 merged to main, so the `@85-move-panel` branch pin is dropped; the bare `cynkra/dockViewR` remote is an interim until cynkra's r-universe rebuilds 0.4.0 (still 0.3.0), then it drops too.
- `dock-grid.R` — `all.equal.dock_grid` drops the transient `focus` marker before comparing. The stored grid carries no focus while the client's echo always does (the active group), which otherwise reads as a spurious geometry difference (the e2e `roundtrip_stable` sentinel caught exactly this). This matches the serdes `drop_focus`; a regression test covers it.
- `utils-dock.R` / `utils-serve.R` — freshen the two "debounced … echo" comments to "settled".
- `test-grid-mirror.R` / `test-model-interleaving.R` — cover the latch: a raised latch skips one echo and a later echo (a real drag, or a server move / resize) commits; the reproduced mirror asserts the restore echo never overwrites the authored grid, then a latch-down echo commits.

### First-class move: surgical `move_panel`

- `panel-ops.R` — `op_move_panel()` relocates the panel in one step via dockViewR's `move_panel(id, position)` (through a new `move_dock_panel()` helper) instead of decomposing into remove + add-with-hint; `hint_to_position()` already produces the `referencePanel` / `direction` vocabulary it consumes. The card rides along (dockview's `moveTo` keeps the panel's content), and the resulting settled echo is committed by the grid mirror — which is what persists the panel's new position.
- `test-panel-ops.R` — assert the single surgical relocation and its idempotency.

### New `resize` verb (#320)

- `dock-view.R` / `panel-ops.R` / `utils-dock.R` — `views$mod` gains a `resize` verb: `resize = list(blk("a", size = 0.3))` sizes a panel's group along its splitview axis through dockViewR's `set_size` (now on the 0.4.0 floor). It mirrors `move` — a placement verb accepting only the `size` hint, resolved through `resolve_placement_verb` and validated (member, ratio in `(0, 1)`). `op_resize_panel()` / `resize_dock_panel()` deliver to the live dock and write nothing to the board directly; the settled echo the resize provokes is committed by the grid mirror, which is what persists the new ratio. #319 had left `resize` rejected as an unknown verb pending the `set_size` proxy.

### Restore the #233 serdes e2e round-trip's grid legs

Reverts the #289 narrowing (`1a03b17`), which dropped the grid-geometry and client-render legs of `test-utils-serve.R`'s Export/Import round-trip because they were flaky against exactly the transient-frame commits this PR removes. With the settled chain plus the restore latch, a restore no longer commits a separate-leaves grid, so the legs are reliable again (dock#327): the fronted-tab `visible_block_ids(active_view_grid(restored))`, the grid round-trip across the reload, the `active_block_panel_tabs(app, "analysis")` client-render assertion, and the `wait_active_block_tabs` / `drop_focus` helpers.

The grid round-trip compares within `grid_size_tol()` rather than byte-for-byte. Under the one-shot restore latch a post-restore `select-panel` echoes the client's rendered sizes, so the stored grid tracks those (jittering sub-tolerance run to run) rather than pinning at the authored value the old blanket server-skip preserved. Structure and membership stay byte-exact; only the sizes are tolerance-compared, the same equality the mirror itself commits on. Byte-exact geometry returns once dockViewR distinguishes a restore in `_state-source` and the mirror can skip the whole restore burst — tracked in #343 (upstream cynkra/dockViewR#89).

Fixes #301
Fixes #320
Fixes #327

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,7 +26,7 @@ Imports:
     shinyjs,
     bslib,
     glue,
-    dockViewR (>= 0.2.1),
+    dockViewR (>= 0.3.0),
     htmltools,
     cli,
     shinyWidgets,
@@ -45,7 +45,8 @@ Suggests:
     rmarkdown
 Remotes:
     nbenn/shinytest2@wait-for-app-serving,
-    BristolMyersSquibb/blockr.core
+    BristolMyersSquibb/blockr.core,
+    cynkra/dockViewR@85-move-panel
 Config/testthat/edition: 3
 Collate:
     'action-class.R'
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,7 +26,7 @@ Imports:
     shinyjs,
     bslib,
     glue,
-    dockViewR (>= 0.3.0),
+    dockViewR (>= 0.4.0),
     htmltools,
     cli,
     shinyWidgets,
@@ -46,7 +46,7 @@ Suggests:
 Remotes:
     nbenn/shinytest2@wait-for-app-serving,
     BristolMyersSquibb/blockr.core,
-    cynkra/dockViewR@85-move-panel
+    cynkra/dockViewR
 Config/testthat/edition: 3
 Collate:
     'action-class.R'
--- a/NEWS.md
+++ b/NEWS.md
@@ -86,6 +86,11 @@
   (`add` / `rm` / `move` / `select`), with panels named by the new typed
   references `blk()` / `ext()` instead of wire-id prefixes. **Breaking:** the
   old set-replace membership form is retired.
+* The `views$mod` grammar gains a `resize` verb --
+  `resize = list(blk("a", size = 0.3))` sets a panel's group size along its
+  splitview axis (a ratio in `(0, 1)`), delivered through dockViewR's `set_size`
+  proxy. Like `move` / `select` it is client-owned geometry: pure delivery,
+  captured by the grid mirror, no board write (#320).
 * **Breaking:** an extension's id is now owned by its container (mirroring
   blocks), serving as its single identity everywhere -- the wire panel id, DOM
   handle, module namespace and `ext()` target.
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -323,35 +323,31 @@ freeze_hidden_inputs <- function(board, visibility) {
 }
 
 # The grid mirror's observer, split from manage_dock for testability. It reacts
-# to the view's settled `_state` echo and commits the client layout verbatim in
-# canonical form, only when it differs (beyond the sash-size tolerance) from
-# what is stored -- so a sash drag or tab switch is at most one board commit, a
-# re-echo after quiescence none, and the echo of the mirror's own restore
-# converges without churn (it matches what was pushed). It does not restrict to
-# membership: a panel absent from the view is an inert ghost, pruned at the
-# compose / restore boundary, never by this writer.
-observe_grid_echo <- function(id, dock, board, commit_grid) {
-
-  # Bridge (#301): a server-side stand-in for the gesture-settled `_state`
-  # emission dockViewR does not yet ship. Current dockViewR streams per-frame
-  # states through a sash drag, so without this the mirror would commit per
-  # frame -- an interactive regression versus main's cadence. Unlike the
-  # pre-rework debounce (which raced a second writer and diffed specs), this
-  # mirror is the sole writer, commits verbatim canonical values, and diffs
-  # nothing: the only cost is 250 ms of bounded staleness on a slot whose
-  # readers are all boundaries. Remove once the settled `_state` chain lands.
-  settled <- debounce(reactive(dock$layout()), 250)
+# to the view's settled `_state` echo, ignores the echoes its own restores
+# provoke (`_state-source` == "server"), and commits the client layout verbatim
+# in canonical form, only when it differs (beyond the sash-size tolerance) from
+# what is stored -- so a sash drag or tab switch is at most one board commit and
+# a re-echo after quiescence none. It does not restrict to membership: a panel
+# absent from the view is an inert ghost, pruned at the compose / restore
+# boundary, never by this writer.
+observe_grid_echo <- function(id, dock, board, session, commit_grid) {
 
   observeEvent(
-    settled(),
+    dock$layout(),
     {
+      source <- isolate(session$input[[dock_input("state-source")]])
+
+      if (identical(source, "server")) {
+        return()
+      }
+
       views <- board_views(board$board)
 
       if (!id %in% names(views)) {
         return()
       }
 
-      state <- settled()
+      state <- dock$layout()
 
       if (is.null(state)) {
         return()
@@ -748,7 +744,7 @@ manage_dock <- function(
         update(list(views = list(grid = set_names(list(grid), id))))
       }
 
-      observe_grid_echo(id, dock, board, commit_grid)
+      observe_grid_echo(id, dock, board, session, commit_grid)
     }
 
     if (get_log_level() >= debug_log_level) {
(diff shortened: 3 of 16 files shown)


nbenn:
When this lands — the settled-`_state` chain replacing the bridge debounce — the grid mirror (`observe_grid_echo`) stops committing dockViewR's per-frame intermediate states, so a restore no longer transiently commits a separate-leaves frame (dock#327).

At that point, revert the `#233` serdes e2e narrowing done in #289. That PR dropped the grid-dependent legs of `test-utils-serve.R`'s "a board survives the live Export/Import round-trip (#233)" because they were flaky against exactly this: the committed grid the export serialises, and the rendered DOM, both caught the mirror's transient collapse, and the settle-wait could even time out when the group never formed.

Restore:
- `visible_block_ids(active_view_grid(restored))` == the fronted tab,
- the `ser grids == ser2 grids` byte compare,
- the `active_block_panel_tabs(app, "analysis")` client-render assertion — the server+client composition seam, the piece with real non-redundant value here,
- and the `wait_active_block_tabs` / `drop_focus` helpers they used.

The exact removed assertions and rationale are in dock#327.

Issue #285, opened 2026-07-02
Busy pulse fires on every panel switch, not just on real work


nbenn:
## Summary

The bslib busy pulse fires on **every dock panel switch**. It's configured in `serve` (`R/utils-serve.R`):

```r
useBusyIndicators(spinners = FALSE, pulse = TRUE),
busyIndicatorOptions(pulse_background = "#5e626b", pulse_height = "5px")
```

The pulse is desirable at startup and during real work (block evaluation), but it now flashes on a plain panel switch, which reads as "the app is processing" for what is only layout bookkeeping. It's visual noise on an interaction that should feel free.

## Why it fires

bslib's pulse reacts to **any** server-busy flush. A panel switch is no longer a purely client-side dockview operation — it round-trips to the server:

- `report_visible_observer` (added in #201) reacts to the active view's layout and writes the on-screen block set to core's `visible` channel, which core consumes to gate block evaluation.
- The live-layout fold commits membership changes to `board_layouts` (`update(list(views = list(mod = ...)))`).

Either flush marks Shiny busy, so the pulse shows. It's the busy-indicator surface of the same panel↔board coupling behind #277 / #279 / #281.

## Fix direction

Scope the indicator to real computation rather than every server flush. Options for whoever picks this up:

- Don't let the visibility-report / layout-fold round-trips count as "busy" — e.g. exclude those updates from the busy state, or drive the pulse off block-evaluation work specifically rather than any recalculation.
- Or debounce / lighten the panel-switch round-trip so a front-tab change doesn't trigger a user-visible busy flush (this would also cut the churn behind #279 / #281).

Startup and genuine recompute should still pulse; a panel switch should not.

## Notes

Same root family as #277 / #279 / #281 (panel switches driving server-side reactivity via the #201 visibility report and the layout fold). Independent of the variadic name-or-position work (#275 / #276).

Issue #232, opened 2026-06-17
e2e: view lifecycle (rename, remove, switch active view)


nbenn:
Part of #228.

Add is covered by the multi-view nav test (#189); rename, remove, and switching the active view are not. The view nav and its client binding (`view_nav_ui`, `view-binding.js`) and the reconcile that tracks views are all dock-owned. Drive each through the view-nav UI and assert that both the nav (one labelled entry per view, unique ids, correct active flag — see `read_view_nav` in `helpers.R`) and the rendered dock panels track the change.

Conventions per #228.

PR #306, opened 2026-07-07
Defer off-screen dockViews and client-confirm the render mark

### LINKED ISSUE #304: Startup starves the visible view behind background construction on multi-view boards
Promoting the profiling result from https://github.com/BristolMyersSquibb/blockr.dock/pull/289#issuecomment-4902313632 to its own issue, so the two coupled defects it identifies are tracked independently of #289's card-deferral work.

On a multi-view board at cdex scale (99 blocks, 12 views), the visible view paints dead-last at startup. The #293 stack removed the startup churn that used to mask this (30 active-group cycles / 14 board updates → 1 / 0 on #300), and with the churn gone the underlying starvation is exposed: net visible paint measured at ~145s stacked, ~78s un-stacked (where the churn accidentally re-established visibility early), against ~28s with background construction held off entirely.

| metric | un-stacked (off main) | stacked on #300 | background held off |
| --- | --- | --- | --- |
| active-group cycles | 30 | 1 | — |
| board updates | 14 | 0 | — |
| net visible paint | ~78s | ~145s | ~28s |

Two coupled defects produce the starvation, per the server timeline in the linked comment:

- **Startup initializes and restores every view's dockView, not just the active one.** The multi-view restore burst leaves the dockview active group pointing at an off-screen view, so visibility transiently drops to 0/99 and the two blocks the user is looking at suspend.
- **The `rendered` mark is reported from the tail of the server-side `show_block_panel` loop, before the client's active-group round-trip settles.** The render gate (core's #257/#259 machinery — the gate itself is correct) then releases background construction, which hogs the single R thread for ~2 minutes; the client events that would settle visibility back to 2/99 queue behind it and land only after construction finishes.

## Fix

Both halves, matching the root cause rather than the symptom:

- **Defer off-screen views' dockViews entirely** — initialize and restore only the active view's dock at startup, mirroring #289's card deferral. No off-screen dockView is ever touched to cycle the active group, so the 0/99 transient cannot happen. The #293 semantics make this well-defined: a never-rendered view is a first-class state (authored or `NULL` grid), and its mirror activates on first render.
- **Report `rendered` only once the visible view is client-confirmed settled** — the mark must be derived from the client's report, not inferred from the end of a server-side loop. This is the same interpret-versus-own discipline the layout rework applied elsewhere, applied to the render gate's input.

One observation from the timeline worth a separate look in core rather than scope here: the active-group burst landed only after all 59 background constructions completed, which suggests the one-block-per-tick background yield is not yielding to WebSocket input processing — any long background run would starve any client interaction the same way. If that reproduces in isolation it deserves its own upstream issue.

## Acceptance

On the cdex-scale board, the visible view paints before background construction begins (the ~28s regime measured with background construction held off), and the 0/99 visibility transient no longer occurs at startup.

nbenn:
## Summary

Promotes the two coupled defects profiled in #289 (comment 4902313632) to their fix. On a many-view board (cdex scale: 99 blocks, 12 views) the visible view painted dead-last at startup once the #293 stack removed the churn that had been accidentally re-establishing visibility early.

- **Defer off-screen views' dockViews.** `reconcile_views()` now builds only the active view's dock; every other view's dock is created on first visit, when a later reconcile finds it active — mirroring #289's card deferral. Initialising every view's dockView at startup left the dock active group pointing at an off-screen view (the transient 0/99 visibility drop), which suspended the two blocks the user was looking at.
- **Report `rendered` from the client, not the server loop.** The per-view `initialized` flag now flips on dockViewR's `_restored` echo — the client confirming it applied the restore and the card moves batched with it — rather than at the tail of the server-side `show_block_panel()` loop. Core's render gate (#257/#259) then releases background block construction only once the visible view is actually on screen, instead of before, where it monopolised the single R process for ~2 minutes while the client's settle events queued behind it.
- **`view_data()` no longer blocks on a never-reporting view.** The old all-or-nothing NULL (every view must report a live layout once) can't hold when off-screen docks are never created. A view without a live dock now contributes its board-stored arrangement — what it would restore to — and upgrades to live once its dock reports. Serialization reads the stored slots directly since #273, so nothing downstream regresses.

Also corrects two tests that were already failing on the #289 tip (`the visibility seed reads the active view's open tabs`, `... falls back to all members when unexpressed`): they still asserted the old flat `visible()` id-set after `de19288` moved that channel to the named `required`/`rendered` status. Independent of this change — they fail on `272-defer-offscreen-cards` without it.

New regression tests cover both halves, each verified to fail on the unfixed code. The cdex-scale acceptance (visible view paints before background construction; no 0/99 transient) and the browser-render behaviour stay CI / real-app territory: the dock e2e layer does not render dockview locally, and workflow CI is dormant while the branch is stacked.

Stacked on #289 (`272-defer-offscreen-cards`). Retarget to `main` once the stack lands.

Fixes #304

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,21 @@
 # blockr.dock (development version)
 
+* At startup the board now builds only the active view's dockView;
+  off-screen views' docks are created on first visit rather than all up
+  front (#304), mirroring the card deferral below. Building every view's
+  dockView left the active group pointing at an off-screen view during the
+  startup restore burst, transiently dropping the visible view's blocks and
+  starving first paint. `reconcile_views()` now builds only the active
+  view's dock and defers the rest; a view without a live dock contributes
+  its board-stored grid to `view_data()` rather than blocking it.
+
+* A view's `visible`-axis mark -- the client-confirmed paint blockr.core's
+  background-construction gate waits for -- now fires on dockViewR's
+  `_restored` echo rather than at the tail of the server-side restore loop
+  (#304). It thus tracks the client actually applying the restore, so the
+  gate holds background construction until the visible view is on screen
+  instead of releasing it while the client is still settling.
+
 * At startup the board now builds only the active view's block cards;
   off-screen views' cards are built on first visit rather than all up
   front (#272). `board_ui()` rendered an edit card for every block across
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -89,11 +89,11 @@ board_server_callback <- function(board, update, visibility, ...,
   report_visible_observer(visibility, client_active, docks)
 
   # `view_data` is the live all-views layout, split into a `dock_views` +
-  # `dock_grids` pair. It is NULL until every view has reported its layout once
-  # (the all-or-nothing in live_view_data), so consumers req() it. A view's
-  # live arrangement is read on demand here; it no longer folds back into the
-  # board's slots.
-  view_data <- live_view_data(client_views, docks, client_active)
+  # `dock_grids` pair. Each view is live where its dock has reported, else its
+  # board-stored grid, so a deferred off-screen view is represented without
+  # blocking (see live_view_data). A view's live arrangement is read on demand
+  # here; it no longer folds back into the board's slots.
+  view_data <- live_view_data(client_views, docks, board, client_active)
 
   # Each extension can read the others' server results through the
   # `extensions` environment: one active binding per id, each resolving to
@@ -217,9 +217,10 @@ report_visible_observer <- function(visibility, client_active, docks) {
 
   # Owns the required axis: over the built cards, the active view's front panels
   # go required TRUE and everything else built FALSE. It leaves the visible axis
-  # alone -- the per-view arrange observer writes it once the client has painted
-  # (mark_cards_rendered), so the two touch disjoint slots. `req(layout())`
-  # waits for the client's report (NULL before then).
+  # alone -- the per-view `_restored` observer writes it once the client
+  # confirms it painted the view (mark_cards_rendered), so the two touch
+  # disjoint slots.
+  # `req(layout())` waits for the client's report (NULL before then).
   on_screen <- reactive({
     active <- req(client_active())
     dock <- req(docks[[active]])
@@ -284,21 +285,24 @@ observe_grid_echo <- function(id, dock, board, commit_grid) {
   )
 }
 
-# Returns NULL while any view is still pending — its dock not yet created by
-# reconcile, or its dockview layout not yet reported — so downstream observers
-# `req()` past the initial flush. Reading `docks[[v_id]]` (a reactiveValues)
-# takes a dependency on each view's entry, so this re-evaluates when reconcile
-# creates the dock — whatever the init flush order — rather than stranding on
-# the empty-`docks` early return. The active view comes from `client_active`,
-# not `client_views`. The live all-views layout is returned split into a
-# `dock_views` (membership + names + active) and a `dock_grids` (geometry) --
-# the same shape the board stores.
-live_view_data <- function(client_views, docks, client_active) {
+# The live all-views layout, split into a `dock_views` (membership + names +
+# active) and a `dock_grids` (geometry) pair -- the same shape the board stores.
+# Each view contributes its live dockview grid once its dock has reported, and
+# its board-stored grid otherwise, so an off-screen view whose dock is deferred
+# (never created this session, #304) is represented by what it would restore to
+# rather than blocking every consumer on a layout that never arrives. Reading
+# `docks[[v_id]]` (a reactiveValues) takes a dependency on each view's entry, so
+# a view upgrades from stored to live when reconcile creates its dock and it
+# reports -- whatever the flush order. NULL now only stands for a view briefly
+# in the nav model but not yet on the board (a sub-flush removal transient),
+# which consumers `req()` past. The active view comes from `client_active`, not
+# `client_views`.
+live_view_data <- function(client_views, docks, board, client_active) {
   reactive({
 
     state <- client_views()
 
-    grids <- lapply(names(state), live_view_grid, docks = docks)
+    grids <- lapply(names(state), live_view_grid, docks = docks, board = board)
 
     if (any(lgl_ply(grids, is.null))) {
       return(NULL)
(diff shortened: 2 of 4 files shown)


nbenn:
Pinning core#259's `blockr.dock` revdep at this PR (the merge queue now builds core#259 against `refs/pull/306/head`, via a `deps` block in #259's body) clears the earlier `busy pulse` (#285) failure — the `visible`-format adaptation on this stack works — but leaves one test red against core#259:

```
── Failure ('test-utils-serve.R:496:3'): view lifecycle: switch, rename, remove a view (#232) ──
Expected `docks$id` to have the same values as `c(first, second)`.
Actual:   "First"
Expected: "First", "Second"
Absent:   "Second"
```

So the remaining accommodation belongs here (or on #289). For reference, what core#259 changed about the `visible` write-channel that `board_server()` hands to the board callback:

**Before** it carried a plain character vector of the on-screen block ids: `visible(c("a", "b"))`. **Now** it carries a **named** character vector, block id → status level: `visible(c(a = "rendered", b = "required"))`, where `"required"` = on screen but the panel is not yet arranged into the view, `"rendered"` = the panel has been arranged, and off-screen blocks are simply absent. Core derives two things from it, and nothing else reads the channel:

- **on-screen** = ids at `required ∪ rendered` → feeds the eval/`needed` gate and the per-block render gate exactly as before (a `required` block still evaluates and paints, so the front-end *can* arrange it);
- **ready** = every on-screen block is at `rendered` → releases the background block-construction pass, which is held until then so it can't starve the active view's arrangement (the point of #257).

`NULL` (nothing reported yet) or `gate_visibility = FALSE` still means "the whole board is on screen and ready", unchanged.

So the front-end contract is: write the map, not the id set; seed the active view's blocks at `required` *synchronously* at callback time (exactly as `report_visible` already seeds the id set today, so core's all-visible default doesn't build every block up front at startup); and promote `required → rendered` per block once its panel is actually arranged (`restore_layout` / `show_block_panel`). That promotion is the signal that releases background construction, so it must track *real arrangement*, not result-quiescence.

On the failing assertion itself: it wants two views (`"First"`, `"Second"`) and sees only `"First"`. From core's side that reads like a coupling between this PR's off-screen-dock deferral and the new gating — `"Second"` isn't materialised (or isn't reported into the `visible` map) at the point the test checks — but it could equally be a stale expectation under the deferral, or a view-switch timing race, so I'm flagging it rather than guessing the dock-side fix. You own the dock internals here.

core#259 is otherwise green and is blocked in the merge queue only on this revdep. Once this lands, the `deps` pin in #259's body drops and the revdep tracks dock `main`, which will then carry the adaptation.

nbenn:
Fixed and pushed to `306/head` (73f8543) — a stale test expectation from the deferral, not a core coupling.

`read_view_docks()` reads the server-inserted `.blockr-view-dock` containers. This PR builds only the active view's dock at startup (the off-screen ones defer to first visit), so at load `docks$id` is just `First`; `Second` materialises when the test switches to it. The startup `expect_setequal(docks$id, c(first, second))` predates the deferral.

I moved that "both docks" check to *after* the switch to `Second` — where it now also exercises the lazy first-visit creation — and assert active-only at startup. The later dock reads in `#232` all run post-switch, so they were already fine; this was the one startup read.

It surfaced only on your revdep because `#232` is the single e2e case whose assertion reads server-inserted DOM (present without a dockview client); the rest need the real client, so they pass on CI and fail headless locally — which is exactly why I could confirm `#232` goes green here after the fix.

The contract writeup matches what the stack implements: write the map, seed the active view at `required` synchronously, promote to `rendered` on real arrangement — now the client's `_restored` echo rather than the server-loop tail. A re-run of core#259's revdep against `306/head` should clear.

nbenn:
**Heads-up for the rebase onto #289.** #289 reworked the card-tracking this PR sits on: the `built$ids` env is gone, replaced by a per-block **card lifecycle map**, and the `visible`-channel derivation now hangs off it. The rename, applied across `board-server.R`, `block-ui.R`, `panel-ops.R`, `dock-view.R`:

| old (this PR's base) | new |
|---|---|
| `new_built_cards()` | `new_card_lifecycle()` |
| `dock$built_blocks` (env, `$ids`) | `dock$cards` (env, `$state`) |
| `built` param (`create_view`, `manage_dock`) | `cards` |
| `manage_dock` field `built_blocks = built` | `cards = cards` |
| `built$ids <- union(built$ids, new)` | `cards$state[new] <- "parked"` |
| `built$ids <- setdiff(built$ids, blocks)` | `cards$state <- cards$state[setdiff(names(cards$state), blocks)]` |
| `block_visibility(on_screen, rendered)` (utils-ui.R, removed) | `reconcile_cards(cards, on_screen, arranged)` + `visible_status(cards)` (block-ui.R) |

`$state` maps each built block id to `parked | required | rendered` (a present key ⇔ built card). `visible_status()` drops `parked` — core's off-screen — so what core receives on the channel is byte-identical to before; `parked` is purely the dock's build ledger.

**Where this PR overlaps #289** — both touch `manage_dock` (the tracker field + the `initialized` reactiveVal), `report_visible_observer`, `reconcile_views` / `create_view`, and the seed. On rebase, resolve those hunks to the lifecycle: `report_visible_observer(visible, client_active, docks, cards)` now takes the lifecycle and reconciles it (`reconcile_cards()` then projects via `visible_status()`), and the seed does the same instead of calling `block_visibility()`.

**The `_restored` client-confirm composes for free.** `report_visible_observer` marks `rendered` by reading `dock$initialized()`. So moving the `initialized(TRUE)` flip from the restore-loop tail to the client `_restored` echo — this PR's whole point on the render mark — needs nothing on the visibility side: `rendered` just lands on the truer client-paint signal, which is exactly what you're after.

Net: the dock-deferral itself is orthogonal to the lifecycle (the map is board-wide and threaded through `create_view`'s `cards` arg regardless of *when* a dock is built), so the rebase is mostly the mechanical rename plus reconciling the shared `board-server.R` hunks. Ping me if any of those hunks are ambiguous and I'll sort them out.

nbenn:
**Update — the card-tracking on #289 was simplified further, so ignore the rename table above.** There is no dock-side card store at all now (no `built_blocks` env, no `cards` lifecycle). Per-block visibility state — `parked` / `required` / `rendered` — lives in **one place: the `visible` reactiveVal core already hands the board callback**. A present key is a built card, so the dock reads built-ness straight off the channel (`built_cards(visible)`) and writes it there (`mark_cards_built` / `drop_cards`, and `report_visible_observer` recomputing levels). Core keys its gates on `required`/`rendered`, so a `parked` entry reads as off-screen — no core change; the channel just carries the full map now.

The corrected rename for your rebase:

| this PR's base | final #289 |
|---|---|
| `dock$built_blocks` / `dock$cards` | `dock$visible` (the reactiveVal) |
| `active_dock$built_blocks` / `active_dock$cards` | `active_dock$visible` |
| `built` / `cards` param (`create_view`, `manage_dock`) | `visible` |
| `manage_dock` field `built_blocks = built` | `visible = visible` |
| `new_built_cards()` + the card env | gone — state is in the channel |
| `report_visible_observer(..., docks, cards)` | `report_visible_observer(visible, client_active, docks)` |
| `block_visibility()` (utils-ui.R) | `card_visibility(built, on_screen, arranged)` + `built_cards` / `mark_cards_built` / `drop_cards` (block-ui.R) |

Everything else stands: your `initialized`-on-`_restored` flip still composes for free (`report_visible_observer` marks `rendered` off `dock$initialized()`, so moving that flip just lands `rendered` on the truer client-paint signal), and the dock-deferral is orthogonal to the channel (threaded through `create_view`'s `visible` arg regardless of when a dock is built). Ping me if any shared `board-server.R` hunk is ambiguous.

PR #311, opened 2026-07-07
Route show_panel through the update channel instead of a dock arg

### LINKED ISSUE #308: Retiring `dock` from the extension surface (#265) breaks blockr.dag: show_panel() still needs it, no view_data replacement
## Summary

Retiring the `dock` handle from the extension server surface (#265, "Retire the arrangement fold and active_dock extension surface") **breaks `blockr.dag`**: its node-select observer calls `blockr.dock::show_panel(node, board, dock)`, but extension servers are no longer passed `dock` — they now receive only `view_data`. Result: single-clicking a node in the Workflow/DAG panel throws

```
Error in observe: argument "dock" is missing, with no default
```

`show_panel()` is still exported and still requires the live `dock` (`dock$proxy`), and `view_data` (read-only layout) offers no way to reach it — so there is currently **no supported path for a dock extension to open a block panel**. This affects every app using `new_dag_extension()`.

Note: this supersedes #152, which proposed *renaming* `dock` → `active_dock` (keeping it on the extension surface). What shipped instead **removed** it — so the "extensions that absorb via `...` are unaffected; those binding by name update their signatures" mitigation in #152 doesn't cover `blockr.dag`, which needs the *value*, not just the name.

## Minimal reproducible example

Only `blockr.core` + `blockr.dock` + `blockr.dag`, all current `main`:

```r
library(blockr.core)
library(blockr.dock)
library(blockr.dag)

board <- new_dock_board(
  blocks     = list(data = new_dataset_block("iris")),
  extensions = list(dag = new_dag_extension())
)
serve(board)
```

Open the app and **single-click the one node** in the Workflow panel.

## Observed

```
Warning: Error in observe: argument "dock" is missing, with no default
 85: observe                    blockr.dag/R/srv.R#63
 86: blockr.dock::show_panel     blockr.dock/R/utils-ui.R#128   # panels <- dock_panel_ids(dock$proxy)
 87: dock_panel_ids
 88: dockViewR::get_panels_ids
```

## Root cause

`board_server_callback()` (`R/board-server.R`) passes to each extension server:

```r
list(board = board, update = update, view_data = view_data, actions = triggers, extensions = peers)
```

— no `dock`. But `show_panel()` (`R/utils-ui.R`, exported) is `function(id, board, dock, type = ...)` and dereferences `dock$proxy`. `blockr.dag`'s observer (`R/srv.R:63`) still calls it with the (now-unbound) `dock`, so the promise errors the moment the panel-show fires.

blockr.dock NEWS confirms the change is intentional: *"Dock extensions now receive `view_data` … The `dock` handle (the active-view `active_dock` mirror) is retired from the extension surface at the same time: it is internal now."*

## Why CI didn't catch it

blockr.dock's tests exercise the extension contract with in-repo fixtures; they don't load the real `blockr.dag` package (separate repo). A consumer break in dag passes dock's green CI.

## Impact

Any board with `new_dag_extension()` — node-click-to-open-panel is dead. On the blockr.cloud gallery that's ~12/13 demos (surfaced there first: `Error in observe: argument "dock" is missing`).

## Question / suggested direction

What's the intended path now that `dock` is internal?

1. Restore an extension-facing way to open a panel — e.g. an exported helper that resolves the active dock from the extension context / `view_data`, or re-expose the active dock to extensions; **and/or**
2. Migrate `blockr.dag` to that mechanism.

Either way it needs a coordinated `blockr.dock` + `blockr.dag` change; today's two `main` branches are mutually incompatible for panel-opening extensions.

## Environment

- blockr.core `b8b8649`
- blockr.dock `8dcd0d0`
- blockr.dag  `6852a57`

(reproduced from source at the shas above)

nbenn:
## Summary

- `show_panel()` no longer takes a live `dock` handle. It emits a `views$show_panel` request on the board `update` channel, and a `board_server_callback()` observer realizes it against the internal `active_dock` (the old body, relocated to `open_panel()`). The dock handle stays internal.
- #265 retired the active-view dock handle from the extension server surface, but the exported `show_panel()` still dereferenced `dock$proxy` — so `blockr.dag`'s node-select observer, calling `show_panel(id, board, dock)`, broke with `argument "dock" is missing`, killing node-click-to-open-panel on every board with a DAG extension.
- Routing through `update` restores a supported panel-open path without re-exposing the dock: extensions read layout via `view_data` and open panels via `show_panel()`, both without the handle. No `blockr.core` change is required — core validates only `blocks`/`links`/`stacks` update slots and passes dock-owned ones through.

Migration for extension authors (breaking):

| before | after |
| --- | --- |
| `show_panel(id, board, dock)` | `show_panel(id, board, update)` |

The paired consumer change is in `blockr.dag` (drops its `dock` formal and calls `show_panel(id, board$board, update)`); this PR is pinned as its `blockr.dock` remote until it lands.

Fixes #308

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,14 @@
 # blockr.dock (development version)
 
+* A dock extension opens a panel with `show_panel(id, board, update)`, which
+  now emits the request on the board `update` channel instead of taking a live
+  `dock` handle; the dock realizes it against the active view internally.
+  Retiring the `dock` handle from the extension surface had left `blockr.dag`
+  -- which opens a block's panel on node click -- with no way to reach it,
+  breaking node-click-to-open-panel on every board with a DAG extension.
+  Routing the request through `update` restores that path while keeping the
+  dock internal (#308).
+
 * Renaming a block no longer crashes a board where that block is absent
   from some view -- a block placed in only one of several views, or
   parked in the offcanvas with no panel at all. The per-view rename
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -102,9 +102,11 @@ board_server_callback <- function(board, update, visible, ...,
   # both -- they are consumed on each side (serialization and the edit-block
   # plugin read the returned pair). The gaps are deliberate: `dock`
   # (active_dock) is returned for core's block insert / remove plugin but
-  # withheld from extensions, which read layout via `view_data`; `board` /
-  # `update` are core's own inputs, not echoed back; the peer env stays
-  # internal (core gets the resolved `ext_res`).
+  # withheld from extensions -- they read layout via `view_data` and open
+  # panels via `show_panel()`, which routes the request through `update` to the
+  # show_panel observer rather than handing out the dock; `board` / `update`
+  # are core's own inputs, not echoed back; the peer env stays internal (core
+  # gets the resolved `ext_res`).
   ext_res <- lapply(
     exts,
     extension_server,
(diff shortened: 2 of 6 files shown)


nbenn:
Closing — superseded. This branch routed panel-opening through a bespoke `views$show_panel` update slot, which the real update cycle rejects (`validate_views_delta`: unknown `views` slice key). #318/#319 solves this properly with the panel-op verb grammar (`add`/`rm`/`move`/`select`) plus a delivery observer, so no dock-side change is needed here: the DAG consumer (BristolMyersSquibb/blockr.dag#149) builds a `views` reveal delta from the exported readers (`board_views` / `active_view` / `view_members` / `as_block_panel_id`) and #318's grammar. #308 is addressed by that pairing.

PR #313, opened 2026-07-08
Add server-initiated view relayout via a views$relayout slice

### LINKED ISSUE #312: Server-initiated view re-layout: one-shot push semantics for authored grid writes
Part of #293 — the last unbuilt piece of the ownership design: the server→client geometry path, with the loop-safe semantics the rework reserved for it.

## The gap

In-place re-layout of an existing view through the update lifecycle has no path. The real caller is blockr.assistant's `modify_view` LLM tool ([tools-view.R#L222](https://github.com/BristolMyersSquibb/blockr.assistant/blob/937214cfc4c6/R/tools-view.R#L222), staging via [staging.R#L411](https://github.com/BristolMyersSquibb/blockr.assistant/blob/937214cfc4c6/R/staging.R#L411)), which stages a fused layout into `views$mod` — now rejected by the membership-only validation, so the staged batch aborts.

For the record, this is not a regression of a previously working path: since the board→dock push was removed (#259 era), a `views$mod` layout never reached a live dock — the stored copy was silently reverted by the membership fold, which is one of the dead-write defects listed in #293's problem statement. The rework made the dead write an explicit rejection, deferred per the settled policy (restore semantics, or an error until a real caller exists). The caller now exists.

## Design: a discrete command, not a sync channel

A server-initiated re-layout is teleport semantics — one write, one push, one ack — never a continuous reconciliation:

- **Lifecycle**: a new `views$relayout` slice carrying a fused `dock_layout` per view id. Augment decomposes it with the existing casts — `as_dock_view()` for membership (validated `members ⊆ blocks` as usual), `as_dock_grid()` for geometry — and apply writes both slots. Unlike `views$mod` (membership sets) and `views$grid` (the mirror's echo channel), `relayout` is the authored full-view write.
- **Delivery**: `manage_dock` observes `update()$views$relayout` for its own view id and issues a one-shot `restore_dock` push of the composed payload — the same event-driven pattern the panel-title observer already uses for `update()$blocks$mod`. A view whose dock is deferred or not yet created needs no push at all: the slots are written and first activation applies them (this composes with #304's deferral for free).
- **Ack, not echo-fight**: the push provokes a client `_state`, and the mirror commits it — that echo is the capture of the client's realization of the pushed layout (including any default-placed members), not noise to suppress. One push, one capture commit, converged — the mirror remains the sole continuous writer, and pushes happen only at births and explicit commands.

The loop-safety property belongs in the #296 harness family: one `relayout` produces bounded commits and then quiescence, under the same seeded-interleaving treatment as the other events.

## Assistant migration

The tool change is mechanical: `modify_view` stages into `views$relayout` instead of `views$mod`, payload unchanged (the fused layout from `layout_from_json`). `add_view`, `remove_view`, `set_active_view`, rename, and the `validate_layout` probe are unaffected; the `list_views` read side maps onto `board_views()` / `board_grids()`.

## Acceptance

A `modify_view` on a live view visibly rearranges it in the client; one relayout yields bounded commits followed by quiescence; a relayout of a deferred view applies at its first activation; membership introduced by a relayout is validated against blocks; a serialize after the ack reflects the new arrangement.

nbenn:
## Summary

- Adds a `relayout` slice to the `views` update channel: a fused `dock_layout()` per view id, resolved and split at augment into that view's membership and stored grid, then written to both board slots in one commit by `apply_views_relayout()`. Membership a relayout introduces is validated against the board's blocks, and a view id may not also travel on `add` / `mod` / `rm` in the same delta.
- Delivers the geometry to the live dock as a one-shot push: `manage_dock` observes `update()$views$relayout[[id]]` (the event pattern the panel-title observer already uses for `update()$blocks$mod`) and, after parking the live cards back in the offcanvas, restores the new arrangement and re-homes each card — through an extracted `push_dock_layout()` it now shares with the initial render. A view whose dock is deferred or not yet created gets no push; it applies the slots when it first renders.
- It is a discrete teleport command, not a continuous sync, so one relayout yields bounded board commits and then quiescence. That property is pinned in the #296 model harness as a new `relayout` event across the seeded interleavings plus a dedicated convergence test.
- Stacked on #300 (`295-restore-rebuild`), whose machinery predates the #298 cast rework, so the decompose uses `grid_from_layout()` exactly as `apply_views_add()` does; it adapts to `as_dock_view()` / `as_dock_grid()` when the stack restacks onto the newer #298. CI gates on `branches: main`, so this stays dormant until the stack lands.
- Follow-up, separate PR in blockr.assistant: migrate the `modify_view` tool from `views$mod` (now rejected by the membership-only validation) to `views$relayout`, payload unchanged.

This is the server-to-client geometry path — the last unbuilt piece of the ownership design (#293).

Fixes #312

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,17 @@
 # blockr.dock (development version)
 
+* Server code can now re-lay-out an existing view in one authored command,
+  through a new `relayout` slice on the `views` update channel (#312). The
+  slice carries a fused `dock_layout()` per view id; the update lifecycle
+  splits it into that view's membership and stored grid, and the live dock
+  is teleported to the new arrangement with a single push. This is the
+  server-to-client geometry path of the structure/grid split (#293) --
+  membership introduced by a relayout is validated against the board's
+  blocks, and it is a discrete command rather than a continuous sync, so
+  one relayout yields bounded board commits and then quiescence. A view
+  whose dock is deferred or not yet created applies the new arrangement
+  when it first renders.
+
 * A board's per-view layout is now split into two independent slots: a
   `dock_views` collection of structure objects (each view's ordered
   panel-id set, name and id, plus the active view), read with
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -508,6 +508,47 @@ reconcile_views <- function(board, update, docks, active_dock,
   invisible()
 }
 
+# Push a whole-view layout onto a live dock: restore the dockview geometry,
+# then re-home each panel's block / extension card into its group. The caller
+# must have parked the current cards in the offcanvas first (as the initial
+# render does), so the dockview rebuild does not tear live UI down with the old
+# panels. Shared by the initial restore and the server-authored relayout.
+push_dock_layout <- function(layout, dock, blocks, extensions) {
+
+  restore_layout(layout, dock$proxy, blocks = blocks, extensions = extensions)
+
+  for (pid in as_dock_panel_id(layout)) {
+    if (is_block_panel_id(pid)) {
+      show_block_panel(pid, add_panel = FALSE, dock = dock)
+    } else if (is_ext_panel_id(pid)) {
+      show_ext_panel(pid, add_panel = FALSE, dock = dock)
+    } else {
+      blockr_abort(
+        "Unknown panel type {class(pid)}.",
+        class = "dock_panel_invalid"
+      )
+    }
+  }
+
+  invisible()
+}
+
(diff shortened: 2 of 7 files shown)


nbenn:
Closing as superseded by #316 — the `views$relayout` lifecycle slice this implements is withdrawn in favor of atomic panel ops with placement hints, which need no new lifecycle verb and leave the mirror as the sole grid writer. The delivery and reducer test patterns here (the park/push mechanics and the mocked-binding `manage_dock` tests) are cited in #316 as prior art for the action handlers' coverage.

PR #317, opened 2026-07-08
Add panel placement ops to the action registry

### LINKED ISSUE #316: Panel placement ops with hints
Part of #293 — supersedes #312 / #313, whose `views$relayout` design conflated two different problems. Adding a panel to a view is a membership event, not a layouting one; what the server-side paths actually lack is a **placement hint**. And a hint is a message, not state: it rides the op, is consumed once at insertion, and is never stored, validated, or reconciled.

## Design: an op algebra, no new lifecycle verbs

- **Add panel to view, with hint** — the mechanism exists end-to-end: `show_block_panel(pid, add_panel = <position>, dock)` already takes dockview's position (`referencePanel` / `referenceGroup` plus `direction` in `within` / `left` / `right` / `above` / `below`); the add-panel modal uses it and `determine_panel_pos()` computes the default. The gap is plumbing only: expose an add-with-hint op through the action registry so extensions and API callers can supply the position. The membership fold captures the set change; the mirror captures where dockview actually placed it.
- **Remove panel from view** — the hide path, exposed through the same registry.
- **Select / activate a panel** — `select_panel`, same treatment.
- **Resize** — `set_size`, joining the family once #303 lands its dockViewR floor; op down, echo up, mirror records, no board write.
- **Move an existing panel** — the one genuine gap in the op family, filed upstream as cynkra/dockViewR#85 (`movePanel` proxy). Interim approximation: remove + add-with-hint.
- **Replace a whole layout** — deliberately not built. Import/restore already has rebuild semantics, and every realistic server-side intent decomposes into the ops above; a bulk primitive waits for a demonstrated bulk need.

No lifecycle changes: `views$mod` stays membership-only, `views$grid` stays the mirror's capture channel, and the ops leave no state behind — a server-initiated rearrangement is indistinguishable, to the board, from the user having dragged the panels there. The #296 harness needs no new machinery for the same reason.

Prior art: #313's delivery and reducer test patterns (park/push mechanics, mocked-binding manage_dock tests) are worth mining for the action handlers' test coverage, even though its lifecycle half is withdrawn.

The consumer migration (retiring blockr.assistant's `modify_view` read-edit-rewrite tool in favor of atomic tools mapping one-to-one onto these ops) is tracked in blockr.assistant.

## Acceptance

An extension or API caller can add a panel to a live view at a stated position and dockview honors the hint; remove, select and resize work through the same registry; the op path writes nothing to the board (fold and mirror capture afterward, one commit each, then quiescence); a hint never appears in any stored or serialized artifact.

nbenn:
## Summary

- Adds `add_panel_action`, `remove_panel_action` and `select_panel_action` to the board's action registry, so an extension or API caller can place a panel in a live view at a stated dockView position, remove one, or activate one -- the same paths the add-panel modal and a tab drag already drive. Each is fired as `actions[["add_panel_action"]](list(panel = , position = , view = ))`; `position` is a dockView hint (`referenceGroup` / `referencePanel` plus a `within` / `left` / `right` / `above` / `below` direction) and `view` defaults to the active view.
- The ops write nothing to the board. They mutate the live dock only, and the view's membership fold plus the settled grid mirror capture the result afterwards -- a server-initiated placement is indistinguishable, to the board, from a hand-arranged one, and the position hint is consumed once at insertion and never stored or serialized.
- The add-panel modal, the tab-close path and layout restore now share the same add / remove / select dispatch helpers the ops use, so a user gesture and a server op run one code path (no behaviour change to the existing paths).
- Built on `main`: add / remove / select need only mechanisms already there (`show_block_panel(add_panel = <position>)`, the hide path, `select_block_panel`). Resize (`set_size`) joins the family once #303 lands its dockViewR floor; a first-class move awaits cynkra/dockViewR#85 and meanwhile decomposes into remove-then-add.

Fixes #316

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,6 +53,7 @@ Collate:
     'action-class.R'
     'action-block.R'
     'action-link.R'
+    'action-panel.R'
     'action-stack.R'
     'action-ui.R'
     'action-utils.R'
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,19 @@
 # blockr.dock (development version)
 
+* Panel placement is now available as ops through the action registry, so
+  an extension or API caller can add a panel to a live view at a chosen
+  position, remove one, or select one -- the same paths the add-panel
+  modal and a tab drag already drive. The add op carries a dockView
+  position hint (`referenceGroup` / `referencePanel` plus a `within` /
+  `left` / `right` / `above` / `below` direction), consumed once at
+  insertion and never stored. Like a user gesture, the op writes nothing
+  to the board: the view's membership fold and the settled grid mirror
+  capture the result afterwards, so a server-initiated placement is
+  indistinguishable from a hand-arranged one. Resize (needs the
+  `set_size` dock proxy) and a first-class move join the family once
+  their dockViewR floors land; a move meanwhile decomposes into
+  remove-then-add (#316).
+
 * The block status badge is now derived in one exported helper,
   `block_status_badge()`, and reused by blockr.dag, so the dock card icon
   and the DAG node badge always render the same status, colour and
--- a/R/action-class.R
+++ b/R/action-class.R
@@ -89,7 +89,10 @@ board_actions.dock_board <- function(x, ...) {
     remove_link_action,
     add_stack_action,
     edit_stack_action,
-    remove_stack_action
+    remove_stack_action,
+    add_panel_action,
+    remove_panel_action,
+    select_panel_action
   )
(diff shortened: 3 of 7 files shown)


nbenn:
Closing as superseded by #318: the action-registry mechanism this implements is withdrawn — panel operations land as nested verbs in the views delta instead (staging-, validation- and batch-compatible by construction). Nothing carries over structurally; the delivery-observer pattern in #318 follows the existing set_panel_title precedent.

Issue #217, opened 2026-06-10
Update board_layouts synchronously instead of via the debounced live-sync


nbenn:
`board_layouts()` lags the live dock, and #204 works around that rather than fixing it. These updates should be synchronous.

Today a block add reaches the live dock immediately through `insert_block_ui.dock_board()`, but only reaches `board_layouts()` later, through the *debounced* live→board sync (`sync_layouts_to_board()` / `layouts_to_board_observer()`, `millis = 250`). Meanwhile `reconcile_views()` drives the board→live direction on every board change. So while `board_layouts()` is stale, a board change fires `reconcile_views()`, which restores the live dock to the stale value and wipes the just-added panels — that is #196.

#204 fixes #196 by making `reconcile_views()` *tolerant* of the lag: it tracks the last committed layout per view (`committed_layouts`) and only restores when that committed layout genuinely changed. This keeps the lagging bidirectional sync in place and papers over it; as more state flows through this path it will get messy.

Goal: a structural change to the live dock (a panel added or removed) updates `board_layouts()` in the same flush, not via a debounced echo. Then the live dock and `board_layouts()` never diverge transiently, `reconcile_views()` never has a stale target to fight, and the `committed_layouts` workaround from #204 can be dropped (reverting `reconcile_views()` to the simpler "restore whenever live ≠ board_layouts").

Constraints to keep in mind:

- The debounce is there because drag-resize emits many geometry events per gesture. Keep debouncing *geometry* (sizes / ratios); apply *structure* (which panels exist, and in which view) synchronously.
- Panel placement on a block add currently has two owners — `insert_block_ui.dock_board()` (incremental live add) and `reconcile_views()` → `apply_layout_diff()` (full restore from `board_layouts()`). A synchronous design must pick one so they don't double-add; the natural split is to let `insert_block_ui()` keep doing the live placement and have the same board-update step record it in `board_layouts()`.
- I could not confirm that `board_layouts()` captures added-block panels at all today — the debounced sync did not appear to fire on programmatic adds in the #196 reproduction. If it doesn't, serialization / export already drops added blocks from a view's layout, so this would also close a persistence gap. Worth confirming as part of the work.

Related: #196, #204.

PR #319, opened 2026-07-08
Route panel operations through the views-delta grammar

### LINKED ISSUE #318: Panel operations as views-delta verbs: add, remove, move, resize, select
Part of #293 — supersedes #316 / #317 (and before them #312 / #313). Those designs routed panel operations around the update lifecycle (a lifecycle slice carrying whole layouts, then the action registry, then session-resolved imperatives). All withdrawn for the same reason: the update lifecycle is the extension API — it is the staging, validation and atomicity boundary extensions and plugins already speak, so panel operations belong *in its payload grammar*, not beside it.

## The grammar

Two nesting levels: outer verbs operate on the views collection (all four already exist), inner verbs operate on panels within a view, under `views$mod$<view-id>`:

```r
update(list(views = list(
  # outer - the views collection
  add    = list(<id or name> = <dock_view / seed layout>),
  rm     = c(<view-id>),
  rename = list(<view-id> = <name>),
  active = <view-id>,
  # inner - panels within a view
  mod = list(
    <view-id> = list(
      add    = list(<panel-id> = list(near = <panel-id>, side = <dir>, size = <ratio>)),
      rm     = c(<panel-id>),
      move   = list(<panel-id> = list(near = <panel-id>, side = <dir>)),
      resize = list(<panel-id> = <ratio>),
      select = <panel-id>
    )
  )
)))
```

`side` speaks dockview's drop vocabulary: `within` / `left` / `right` / `above` / `below`. Splitting a tab group is not a verb — a directional `move` splits (dockview creates the new group), `within` joins; dockview owns all tree restructuring and the mirror captures whatever it settles to. Groups have no names anywhere in this grammar, deliberately: the canonical grid stores no group ids, so every group-touching op addresses its group through a member panel ("the group containing X"), matching dockview's own `referencePanel` idiom — this is what `resize` keys on.

## Verb semantics

| verb | augment / validate | apply (slots) | delivery (event-keyed observer on the applied payload) | capture |
| --- | --- | --- | --- | --- |
| inner `add` | ids resolve; panel ∈ blocks/exts; not already a member; hint refs resolve | members ∪ panel — the hint is never stored | `add_panel` at the hint's position, else `determine_panel_pos()` | fold no-ops (setequal); mirror records the placement |
| inner `rm` | member exists | members ∖ panel | hide + remove panel | ghost in the grid is a legal transient; next echo clears |
| `move` | panel is a member; hint resolves | nothing — geometry is client-owned | move op (cynkra/dockViewR#85; interim remove + add-with-hint, at the cost of panel re-creation) | mirror captures the new grid |
| `resize` | panel is a member; ratio in (0, 1) | nothing | `set_size` on the panel's group (available once #303 lands its dockViewR floor) | mirror captures the settled ratios |
| `select` | panel is a member of that view | nothing | select the tab — never switches views | mirror captures the active tab |

Rules that hold it together: inner `add` / `rm` are the only slot writers; `move` / `resize` / `select` are pure payload messages — validated at augment, applied as ops, captured by the mirror, never written, never reconciled. Application order within a batch is `rm → add → move → resize → select`, with references validated against post-add membership. Per-view, per-verb maps merge by key, so staged batches (the assistant's pending model) compose instead of clobbering. Delivery observers follow the `set_panel_title` precedent — keyed on the applied `update()` payload, never on state diffs.

Cross-view reveal is composition, not a verb: `active` (outer) plus `select` (inner) in one batch. Convenience comes from pure payload builders — `reveal_panel(board, panel)` returns that composed delta — keeping the grammar primitive. This restores the supported path #308 lost (blockr.dag's node click becomes `update(reveal_panel(board, node))`); #308 closes once blockr.dag adopts it.

## Completeness and testing

The inner verb set is exactly the user gesture set — the five drop targets, the sash drag, the tab click, the close button, the add modal — so anything a user can produce by direct manipulation, an extension can produce in one batch. That equivalence is also the test oracle for the #296-family harness: drive the same arrangement once by ops and once by (simulated) gestures, and the captured grids must agree.

Consumer migration: blockr.assistant's atomic tools (BristolMyersSquibb/blockr.assistant#64) become stagers of these payload entries. Set-replace `views$mod` retires in favor of the add/rm form — the membership fold already computes add/rm via `setdiff` before building a set, and the block-removal cascade is an `rm`.

## Id currency and typed refs

The `block_panel-` / `ext_panel-` prefix strings are demoted to what they really are — the wire and serialization encoding — and leave the API surface. External consumers deal only in block and extension ids.

The public vocabulary is a pair of typed refs, `blk()` and `ext()`: thin exported constructors over the existing classed panel-id machinery (`as.character()` on a ref yields the canonical encoding — the constructor is the codec, nobody hand-writes a prefix). Each ref optionally absorbs its own placement hint, which makes the verb maps unnamed lists of self-describing references and removes the string-key constraint entirely:

```r
update(list(views = list(mod = list(
  v1 = list(
    add    = list(blk("a", near = "b", side = "right"), ext("dag")),
    rm     = "d",
    move   = list(blk("e", near = "a", side = "within")),
    resize = list(blk("a", size = 0.4)),   # once the set_size floor lands (#320)
    select = "a"
  )
))))
```

Rules:

- **Hint fields are valid only where placement happens.** `add` consumes `near` / `side` / `size`; `move` consumes `near` / `side`; `resize` consumes `size` only (unifying resize as `blk(id, size = )`). A hinted ref on `rm`, on `select`, as a `near` anchor, or inside the authoring DSL is a loud error — in `dock_grid()` / `panels()` the arrangement is the structure, so refs there are hint-free by definition.
- **Bare strings are sugar** everywhere refs appear, resolved block-first with a hard error on a true clash (the id exists in both namespaces) that demands a typed ref. Consequences: today's bare extension-alias seeds keep resolving (no block by that name), and a bare ref's meaning can only transition valid → error when a clash later appears, never valid → different target — no silent retargeting is possible. `near` takes a ref or a bare string, never a nested hint.
- **Duplicates are loud**: the same panel twice in one batch with conflicting hints is a caller error, not a last-wins race. Staging merges by concatenation; augment dedups and errors on conflict.
- **Unknown hint keys cannot exist**: hints are constructor arguments, so `blk("a", sizee = 0.3)` fails at the call site with R's own unused-argument error before a payload is ever built. The constructor signature is the validation.
- **Reads stay canonical.** `view_members()` and grid leaves keep the encoded form because type must survive in-band on the way out; `panel_obj_ids()` and the type predicates remain the consumer helpers. Augment canonicalizes refs and extracts hints for the delivery observers; slots and serialization are untouched.

Consumer effect: blockr.assistant's tools (BristolMyersSquibb/blockr.assistant#64) construct refs server-side from flat `{id, near, side}` tool arguments — the schema shape models handle reliably — and never see a prefix.

nbenn:
## Summary

- Panel operations move into the update lifecycle's `views$mod` payload as per-view verbs, so an extension or plugin rearranges a view's panels through the same staging, validation and atomicity boundary every other board change already speaks — rather than beside it (the reason #312/#313 and #316/#317 were withdrawn). `add` and `rm` write membership; `move` and `select` write nothing and exist only as client ops captured by the settled-echo grid mirror. Placement rides `add` / `move` as a `near` anchor + a `side` in dockView's drop vocabulary.
- Panels are named by **typed refs** — `blk(id)` / `ext(id)` — never the `block_panel-` / `ext_panel-` wire prefix. A ref carries its own placement hint (`blk("a", near = "b", side = "right")`), so a verb's operands are an unnamed list of self-describing refs; a bare id string is accepted as sugar, resolved block-first with a loud error on a true cross-namespace clash. Augment resolves refs and sugar to the internal canonical hint form, idempotently — the form the gestures and the block-removal cascade already emit passes straight through. This currency lands here, not a cycle later, because shipping a prefix-string payload only to change it after would repeat the churn this lineage exists to end. The same references also name panels in the layout-authoring DSL (`dock_grid()`, `panels()`, a `views` entry), so a board definition reads as blocks and extensions rather than wire ids.
- **Breaking — extension identity is container-owned** (mirroring blocks): `new_dock_board(extensions = )` names an unnamed extension by its class stripped of the `_extension` suffix (`new_dag_extension()` → `dag`), an explicit list name overrides, and duplicate keys error rather than silently colliding. That key is the single identity everywhere — the wire panel id (`ext_panel-dag`, was `ext_panel-dag_extension`), the DOM handle, the Shiny module namespace, `dock_ext_ids()`, and what `ext()` resolves against — so `extension_ui()` / `extension_server()` take the key from the container just as `block_ui` does; `extension_id()` stays the class, used only for the default-key derivation and the `_extension` subclass check. This is what `ext()` needed: before it, `ext("dag")` produced `ext_panel-dag` while the real wire id was `ext_panel-dag_extension`. Serialization preserves the key so aliased extensions round-trip; a board saved under the old class-derived ids restores with the stale extension panel pruned rather than erroring.
- **Breaking:** the old set-replace form (`views$mod$<view-id>` = a bare membership vector) is retired in favour of the `add` / `rm` verb form the gestures and the block-removal cascade emit; per-view, per-verb maps merge by key so staged batches compose instead of clobbering.
- A per-view apply observer is the **sole mutator** of the live dock, keyed on the applied `views$mod` (the `set_panel_title` precedent), idempotent against the live panel set so a re-augmented payload settles without ping-pong. A panel whose block is removed in the same update is left to core's block-removal path in the active dock (delivering it here too would double-remove and throw client-side); inactive docks apply the cascade `rm` themselves.
- Every membership change reaches the board through `update()` at its source: the add-panel modal and the tab close emit `update(mod = add/rm)` at the click rather than mutating the dock directly, so membership is authoritative from the update and the old fold — a retro-sync for direct mutations (#217) — is deleted. The grid mirror stays, since geometry genuinely is client-owned.
- The supported way for a dock extension to open a block's panel is the `views` update grammar itself: compose `active` + `select` (with a `mod` `add` first for a panel no view holds yet) and pass it to `update()`, with no live `dock` handle — which is off the extension surface. The exported `show_panel()`, which required that handle and cannot emit the grammar, is removed. #308 closes once blockr.dag adopts the grammar.

Scope: `add` / `rm` / `move` / `select` ship on current floors. A first-class `move` awaits cynkra/dockViewR#85 and meanwhile decomposes into remove + add-with-hint (the interim the issue and blockr.assistant#64 both specify). `resize` needs the `set_size` proxy (unmerged dockViewR#78, pinned only by the still-open #303), so it is not in this PR — the verb is rejected until then, failing loud rather than silently no-op-ing, though `blk(id, size = )` is already a valid ref whose `size` validates as a ratio, so the grammar is stable for when the floor lands (tracked in #320).

Migration:

| before (set-replace) | after (verbs) |
| --- | --- |
| `mod = list(V = c("block_panel-a", "block_panel-b"))` | `mod = list(V = list(add = list(blk("a"), blk("b"))))` |
| grow membership by a set | `add = list(blk(<id>, near = <id>, side = <dir>))` |
| shrink membership by a set | `rm = list(blk(<id>))` — or the bare `"<id>"` sugar |
| — (not expressible) | `move = list(blk(<id>, near = <id>, side = <dir>))` |
| — (not expressible) | `select = blk(<id>)` |

Fixes #318

```deps
BristolMyersSquibb/blockr.dag#149
```

The `deps` block points the reverse-dependency check at blockr.dag#149 (the coupled adaptation to the container-owned extension ids and bundled result delivery) instead of its `main`, since #319 and #149 must land together. Drop it once #149 merges.


R/board-server.R
```diff
@@ -559,30 +559,49 @@ manage_dock <- function(
...
 
-          folded <- fold_live_membership(
+          mod <- fold_live_membership(
             view_members(views[[id]]), live_panels()
           )
 
-          if (!is.null(folded)) {
-            update(list(views = list(mod = set_names(list(folded), id))))
+          if (!is.null(mod)) {
+            update(list(views = list(mod = set_names(list(mod), id))))
```
nbenn: Why do we route internal modifications through the `update` api? Would we not rather use non-API functions for this?

nbenn:
Reviewed against #318. Verdict: a faithful implementation of the grammar with the right instincts — `resize` loudly rejected until the `set_size` floor with #320 tracking it, `move` decomposing into remove + add-with-hint as specced (card parked so server state survives), the fold converted to emit `add` / `rm`, set-replace retired with a migration table, delivery idempotent against the authoritative live panel set in `rm -> add -> move -> select` order, and `reveal_panel()` as a pure builder with active-first holding-view resolution. What is missing clusters almost entirely around one dimension the test suite never touches: views other than the active one. Five findings, ranked — the first is the one to fix before undrafting.

**1. Inactive-view delivery steals the single-instance card.** The per-view delivery observer runs in every view's `manage_dock` with no active-awareness, and `op_add_panel` -> `show_block_panel` -> `show_block_ui` (`utils-ui.R`) moves the block's one card into the target dock. Adding block b to inactive view B while b is visible in active view A physically removes b's card from A until the next view switch re-parks it. Nothing in the tests exercises an inactive view, and "stage: add the table to view 2" is a day-one assistant use case. Suggested direction: delivery to a non-active dock performs the dockview panel op but leaves the card parked (activation already un-parks via `show_view_ui`), or defers entirely to activation — either way, with inactive-view tests.

**2. The cascade-`rm` skip is global, but its own justification only covers the active dock.** Panels whose block is removed in the same update are skipped everywhere because "core's block-removal path owns the active dock" — which is exactly the problem: core cleans only the active dock, so a removed block's panel lingers in every non-active live dock (its card and server UI already destroyed) until some boundary. Pre-existing latent bug, but this PR is positioned to fix it rather than codify it: scope the skip to the active view and let the cascade's `rm` entries deliver to the others.

**3. Same-batch `near` forward references validate but deliver dangling.** Validation anchors against post-add membership (spec-correct), but delivery walks the `add` map in payload order — `add(a, near = c)` ahead of c's own add sends dockview a `referencePanel` that does not exist yet. Either dependency-order the adds at delivery (fixpoint pass, cycle -> validation error) or tighten validation to pre-existing members; test whichever.

**4. Unknown hint keys pass silently.** The hint validator checks `near` and `side` but never rejects extras — so `size = 0.3`, which #318's grammar documents, validates fine and is silently dropped at delivery. Inconsistent with this PR's own loud-rejection stance for `resize`; an extra-key abort with the `size` message pointing at #320 is three lines.

**5. `reveal_panel()` returns `NULL` for unplaced panels — quietly re-breaking the #308 case it exists for.** The DAG renders all blocks including unplaced ones; a node click on a picker block gets `NULL` and nothing happens. Either compose the unplaced case (inner `add` to the active view with default placement plus `select`, one batch) or bless the `NULL` contract explicitly and have blockr.dag handle it — but decide, rather than defaulting into silence.

Housekeeping: #311 is a live leftover from the superseded design era, competing with `reveal_panel` — it should close as superseded by this PR. And the gesture-equivalence oracle (op-sequence and gesture-sequence produce identical captured grids) belongs in #299's port scope so it does not fall between the cracks.

R/board-server.R
```diff
@@ -559,30 +559,49 @@ manage_dock <- function(
...
+    # The inverse of the fold: deliver server-initiated panel ops to this view's
+    # live dock. Keyed on the applied `views$mod` for the view (the
+    # `set_panel_title` precedent), never on a board diff — `move` / `select`
+    # write nothing to the board, so a state-diff observer would never see them.
+    # Each op is idempotent against the live panel set, so the fold's own
+    # capture echo (an add / rm this dock already reflects) settles without
+    # ping-pong.
+    observeEvent(
+      update()$views$mod[[id]],
+      deliver_panel_ops(
```
nbenn: Why "deliver"? I'd prefer something like "apply" panel ops.

R/utils-ui.R
```diff
@@ -111,18 +111,84 @@ determine_panel_pos <- function(dock) {
...
+#' @examples
+#' brd <- new_dock_board(
+#'   blocks = c(a = blockr.core::new_dataset_block()),
+#'   views = list(one = "a", two = "a")
+#' )
+#' reveal_panel(brd, "a")
+#'
+#' @rdname panel
+#' @export
+reveal_panel <- function(board, panel) {
```
nbenn: I don't think we need to export this function. Our API provides all that is needed for consumers here to implement this themselves. And as such, I believe we can scratch this function entirely, no?

R/utils-ui.R
```diff
@@ -111,18 +111,84 @@ determine_panel_pos <- function(dock) {
...
+
+  blockr_abort(
+    "Cannot reveal {panel}: no such panel, block or extension.",
+    class = "dock_reveal_panel_unknown"
+  )
+}
+
 #' @rdname panel
 #' @export
 show_panel <- function(id, board, dock, type = c("block", "extension")) {
```
nbenn: This function as well can probably be deprecated, no? This is superseded by the `update` API I believe.

nbenn:
Correction to finding 1 of my review, after verifying `inst/assets/js/show-hide-block.js`: the card steal I described cannot happen. The `move-element` handler explicitly drops moves whose target sits in an inactive view dock, and `show_view_ui` re-issues the move for every panel of a dock at activation — so an add to an inactive view leaves the card exactly where it is, adds the (empty-mount) dockview panel, and the card arrives on the next switch. Inactive-view adds therefore work end-to-end today.

The residual ask shrinks accordingly: (a) tests covering the inactive-view path (panel present in the target dock, card untouched in the active one, card arrives at activation), since the behaviour currently rests on an undocumented client-side guard; (b) a one-line comment on `op_add_panel` noting that the card half is active-guarded in the JS, so that guard is never 'cleaned up' as dead code. Findings 2-5 stand unchanged — in particular finding 2 is unaffected, since it concerns the dockview panel lingering in non-active docks on block removal, not the card.

nbenn:
Thanks for the thorough read — pushed 11ffb2a addressing all five.

1. Delivery is now active-aware: an inactive view runs the dockView panel op only (`add_*_panel` / `remove_*_panel`) and never moves the single board-level card, so staging an add into an inactive view no longer yanks the card off the active one — activation re-parks it via `show_view_ui`. Covered by `op_add` / `op_remove` tests on the inactive path.
2. The cascade-`rm` skip now scopes to the active view: inactive docks deliver the `rm` to clear the wrapper core never reaches, so this fixes the lingering ghost rather than codifying it, and only the active dock defers to core.
3. Add-hint `near` anchors against pre-existing members (`post_rm`) only, so a same-batch forward reference is rejected at validation rather than delivering a dangling `referencePanel`. I went with tightening validation over dependency-ordering the delivery — simpler, and a stager can order its own batch.
4. `validate_panel_hint` rejects unknown keys; `size` gets a message pointing at #320.
5. `reveal_panel()` composes an inner `add` (default placement) + `select` into the active view for an unplaced panel instead of returning `NULL`, so a node click on a picker block reveals it; `NULL` now only for the degenerate no-active-view case.

Coverage: added the `validate_view_mod` branch assertions and the inactive-view delivery tests. Housekeeping: #311 is already closed and its competing `open_panel` / `views$show_panel` path never merged (no such code on `main`), so there is nothing to remove; noted the gesture-equivalence oracle for #299's port scope.

(Will reply to the fold-through-`update` thread once that review lands — short version: `update()` is the only board-write channel the dock module has, and it's the same one the grid mirror's `commit_grid` already uses.)

R/board-server.R
nbenn: `update()` is the only board-write channel the dock module has — core owns `rv$board` and mutates it solely in the apply phase, so any callback or plugin (and this internal fold) reaches the board through `update()`. It's also the same channel the grid mirror's `commit_grid()` uses right below (`update(list(views = list(grid = ...)))`), both from the single-writer layout-ownership model, and routing the membership write through it earns validation plus atomic composition with a concurrent block-removal cascade (`merge_views_mod`) in the same flush. A non-API path would need core to expose a direct board-membership setter — bypassing the single `apply_board_update` reducer and that composition, and it would touch the mirror identically. Happy to pursue that as a core-side follow-up if you'd rather keep internal captures off the public channel; leaving this open for that call.

R/board-server.R
nbenn: Renamed to `apply_panel_ops()` (and the `deliver` wording) in 744f1ee.

R/utils-ui.R
nbenn: Agreed — removed `reveal_panel()` and `resolve_reveal_panel()` in 744f1ee. A dock extension now composes the delta directly, `update(list(views = list(active = v, mod = list(v = list(select = p)))))`, with a `mod` `add` first for an unplaced panel. This also settles finding 5 from your verdict via the "consumer handles it" option — the resolution policy (which view to reveal in) was arguably consumer-specific anyway. #308's path is the grammar itself; blockr.dag adopts it in its own PR.

R/utils-ui.R
nbenn: Deprecated `show_panel()` in 744f1ee — a `.Deprecated()` pointing at the update grammar, kept exported so blockr.dag's existing call does not hard-error before it migrates. Given the forward-only norm and that it is already non-functional for extensions (no `dock` on the surface), I am happy to remove it outright instead — leaving this open for your call on deprecate vs. remove.

R/board-server.R
nbenn: Deleted the fold entirely. The two live gestures that change a view's membership — the add-panel modal and the tab close — were all it captured, and both are server handlers we own; they now emit `update(mod = add/rm)` at the click, so the apply observer is the sole dock mutator and membership is authoritative from the update, with no retro-sync. There's no longer any internal mutation routed through `update()` for the fold to reconstruct — the mutation *is* the update. (The grid mirror stays: geometry genuinely is client-owned, with no server handler behind a drag.)

Landed in e17a8da; this is what your question and this thread were both pointing at, and you were right.

R/utils-ui.R
nbenn: Removed it entirely rather than deprecating. Deleting the fold leaves no way to keep `show_panel()` functional: its add-a-non-member branch would desync board membership without the fold, and as an exported util taking the live `dock` it can't emit the grammar. So removal settles the deprecate-vs-remove question here — the extension-facing path is now the update grammar, full stop.

nbenn:
Addendum to the review, from the id-currency design just settled on #318 (new "Id currency and typed refs" section there — the spec of record): please land it in this PR alongside the findings, since shipping a panel-id-string payload API for one cycle only to change its currency after would repeat the churn this lineage exists to end.

Concretely, on top of the five findings:

- **Add the `blk()` / `ext()` typed refs** — thin constructors over the existing classed panel-id machinery, each optionally absorbing its placement hint (`near`, `side`, `size`). The verb maps become unnamed lists of self-describing refs; the named-list hint maps go away.
- **Resolution moves into augment**: bare strings accepted everywhere refs appear, block-first, hard error on a true clash demanding a typed ref. `rm` / `move` / `select` / `near` resolve against the view's membership; `add` against post-state blocks and extensions.
- **Hint-field validity is contextual and loud**: `add` consumes `near` / `side` / `size`, `move` consumes `near` / `side`, `rm` / `select` / `near`-anchors / the authoring DSL reject hinted refs. `resize` stays rejected per #320, but the ref type accepts `size` now so the grammar is stable when the floor lands.
- **Finding 4 resolves structurally**: hints as constructor arguments mean unknown keys fail at the call site (unused argument) — no whitelist needed; drop that approach.
- **Duplicate refs for one panel in a batch with conflicting hints**: loud error at augment; staging merges by concatenation.

Findings 1 (revised per my correction: inactive-path tests + the JS-guard comment), 2, 3 and 5 are unchanged and compose with this — the same augment pass that resolves refs is where the same-batch `near` ordering (finding 3) belongs.

nbenn:
Landed the id-currency redesign alongside the findings — the `block_panel-` / `ext_panel-` prefixes are now wire-only and off the API.

- **`blk()` / `ext()`** (`R/panel-ref.R`): thin exported typed refs over the classed panel-id machinery; `as.character()` on one is the canonical encoding. Each absorbs its own hint, so a verb's operands are an unnamed list of self-describing refs.
- **Resolution moved into augment** (`resolve_views_mod`): bare strings accepted everywhere refs appear, block-first, with a hard error on a true cross-namespace clash demanding a typed ref. `rm` / `move` / `select` / `near` resolve against membership, `add` against post-state. It is **idempotent** — the internal named-hint form (the gestures, the block-removal cascade, a re-augment pass) passes straight through, so the augment loop converges.
- **Hint validity is contextual and loud**: `add` consumes `near` / `side` / `size`, `move` `near` / `side`; a hint on `rm`, `select`, or a `near` anchor errors. **Finding 4 dissolved structurally** — hints are constructor arguments, so a misspelling is R's own unused-argument error; I dropped the runtime whitelist.
- **Duplicates are loud**: the same panel twice in one batch with conflicting hints errors at augment.
- **`resize` stays a rejected verb** per #320, but `blk(id, size = )` is a valid ref and `size` validates as a ratio in (0, 1), so the grammar is stable for when the `set_size` floor lands.
- **Reads stay canonical** (`view_members()`, grid leaves); slots and serialization untouched.

Findings 1–5 compose unchanged — finding 3's same-batch `near` ordering is checked post-resolution against pre-existing membership, in the same augment pass. blockr.assistant#64 can now build refs server-side from its flat `{id, near, side}` tool arguments and never see a prefix.

nbenn:
One more piece of the id-currency design, settling what `ext()` resolves against: extension identity.

Extensions currently carry two identities — `extension_id()` is `class(x)[1L]` ("dag_extension") and doubles as the collection key, while the name given in `new_dock_board(extensions = list(dag = ...))` rides along as an alias attribute, with `panel_id_map()` resolving both. That dual identity is why extension refs feel unsettled: `ext("dag_extension")` is a class name doing an id's job, and `ext("dag")` only works if the app author happened to alias it.

**The rule: for singletons-by-type, identity is the type.** A board typically holds one extension per type, so the id derives from it — and that derivability is the property that matters for consumers who never see the startup script: anyone who knows *which* extension they mean can compute its id from the type alone. The human behind the browser never types an id at all — they see `extension_name()` display labels in the UI or talk to the assistant; ids are consumed by code (which derives by type) and by the LLM (which discovers via its list tools and echoes back). Stable, discoverable, derivable-by-type is the full requirement set, and one scheme delivers all three:

- `new_dock_board(extensions = )` wraps a bare extension in a list and auto-names any unnamed slot as `sub("_extension$", "", class(x)[1L])` — so `new_dag_extension()` gets the key "dag". Explicit list names override, as today.
- **Duplicate keys error loudly** ("two dag extensions; name them in the `extensions` list") — the one case where type-derived identity breaks, fixed by forcing explicit names rather than minting random ones, since derivability is the point.
- A class not ending in `_extension` makes the `sub()` a no-op and the full class name becomes the key — the rule stays total, no special case.
- **The dual identity retires**: the alias attribute goes, `panel_id_map()`'s both-ways mapping collapses to one lookup against the collection keys, `dock_ext_ids()` returns the keys, the wire encodes them (`ext_panel-dag`), and `ext()` resolves against them. One name per extension, everywhere — construction seeds and update payloads resolve identically. The short keys are namespace-safe because `ext()` carries the namespace; the long class-derived form's only virtue was bare-string clash insurance, which the typed refs already retired.
- Migration is a graceful degrade, not a break: existing exports carrying `ext_panel-dag_extension` become unknown members at restore and are dropped by the existing pruning — the view renders without the extension panel rather than erroring. Touchpoints riding along: `visible_exts()`'s default `"dag_extension"` becomes `"dag"`, plus a NEWS line for the key change.

This composes with the earlier addendum: the same augment pass that resolves `blk()` / bare strings resolves `ext()` against these keys.

PR #325, opened 2026-07-12
blockr.dock CRAN release prep: vendor blockr.ui, drop all remotes

### LINKED ISSUE #326: Flaky e2e test: busy-pulse `bookkeeping` probe reads `block` (test-utils-serve.R:801)
`test-utils-serve.R:801` ("busy pulse tracks real work, not layout bookkeeping (#285)") fails intermittently with a busy-pulse scope mismatch. It reddened both `ci / smoke` and `ci / coverage` on a **docs-only** PR (#325 — only `NEWS.md` and `cran-comments.md`), whose base commit `e1b5eee` passed green in the merge queue on 2026-07-10. A docs-only delta cannot change test behaviour, so this is a flake rather than a regression; the `merge_group` run history shows other sporadic `failure`s too.

## Evidence

Run https://github.com/BristolMyersSquibb/blockr.dock/actions/runs/29205239336 — jobs `ci / smoke` (86683431892) and `ci / coverage` (86683431897) both fail on this one test, `[ FAIL 1 | WARN 0 | SKIP 0 | PASS 7288 ]`:

```
── Failure ('test-utils-serve.R:801:3'): busy pulse tracks real work, not layout bookkeeping (#285) ──
Expected `probe$bookkeeping` to be identical to "none".
Differences:
`actual`:   "block"
`expected`: "none"
```

## Mechanism

The test synthesises the busy state in one `get_js()` snippet: it sets `html.shiny-busy`, appends a `.recalculating` marker to `document.body` (outside `.blockr-view-container`), and reads the pulse pseudo-element's computed `display`, expecting `"none"`.

The pulse is hidden by a single scoped rule (`inst/assets/css/blockr-dock.css:1173`):

```css
html[data-shiny-busy-pulse].shiny-busy:not(:has(.blockr-view-container .recalculating))::after {
  display: none;
}
```

So the pulse reads `"none"` only when there is **no** `.recalculating` anywhere inside `.blockr-view-container`. The test's `bookkeeping` probe relies on that emptiness, but `app$wait_for_idle()` waits for the server to go idle — it does not guarantee the client has cleared the `.recalculating` class from every freshly-rendered output inside the container. When a real output is still marked `.recalculating` at probe time, `:has()` matches it, the `:not()` fails, and the pulse stays `display: block` — so `bookkeeping` reads `"block"`. The synthetic marker placed on `document.body` is irrelevant to the match; only in-container elements count.

## Proposed mitigation

Gate the probe on the container actually being quiescent — poll for zero in-container `.recalculating` before sampling, rather than trusting `wait_for_idle()` alone:

```r
app$wait_for_idle()
app$wait_for_js(
  "document.querySelectorAll('.blockr-view-container .recalculating').length === 0"
)
```

That makes the outside-vs-inside synthetic markers the only differentiator at probe time, which is exactly what the two assertions assume. Happy to open a small PR for this.

nbenn:
Prepares blockr.dock for a CRAN release by removing all four `Remotes` pins — on top of the earlier 0.1.2 release-notes trim and the busy-pulse e2e fix.

## Vendoring blockr.ui (the CRAN blocker)

blockr.dock imported its block-browser, link-menu, stack-menu and sidebar components from `blockr.ui`, which is not on CRAN. The full transitive closure — 5 R files (~1.4k LOC) and 8 CSS/JS assets — is now bundled into blockr.dock as internal components, dropping the dependency:

- `htmlDependency` factories rewired to `blockr.dock`; `rlang::abort` / `stats::setNames` swapped to `blockr.core` wrappers, so **no new deps**.
- The blockr.dock ↔ blockr.ui cycle is resolved (`new_dock_stack` becomes a plain internal call).
- Name collisions fixed: vendored `block_card` → `browser_block_card`; `is_hex_color` deduped onto the hardened blockr.dock copy.
- 38 `blockr.ui::` call sites de-qualified; the `local_mocked_bindings` tests retargeted to the blockr.dock namespace.

## Remaining remotes dropped

`blockr.core` 0.1.3 and `dockViewR` 0.3.0 are now on CRAN, and the shinytest2 e2e tests `skip_on_cran`, so all four remotes are gone and the `Remotes:` field is removed entirely. `utils` (previously undeclared) is added to Imports.

## Also on this branch

- The 0.1.2 NEWS section trimmed from 29 verbose entries to concise headlines.
- cran-comments updated.
- The busy-pulse e2e test made hermetic (fixes #326).

Validation: `lint_package()` is clean and the 108 action unit tests pass. CI runs the full `R CMD check --as-cran` plus the e2e menu tests on stock shinytest2 (the real check that dropping the fork is fine). The 19 vendored functions are kept internal — no change to blockr.dock's public API.


DivadNojnarg (approved):
Looks good. Then we'll move again the sidebar in blockr ui, but the constructions of blocks and links and cie, will stay here.

Issue #331, opened 2026-07-15
Off-screen views' cards built without the served ctrl_block plugin → control toggle / AI sparkle missing


christophsax:
## Symptom

The per-block **control toggle** (core's `ctrl_block` plugin — surfaced as the AI "sparkle" when the served plugin is `blockr.ai::ai_ctrl_block`) shows on blocks in the **initially-active view** but is **absent** on blocks in every **other view** (whose cards are built on first visit). Same block type, same `external_ctrl`.

On the CDEx board: `Treatment Emergent AE` (table, active view) **has** it; `Pre-rand` / `Post-rand Disposition` (tables, on their own view tabs) **don't**. This is **view-based, not block-type-based** — an earlier "named vs `TRUE` `external_ctrl`" theory was a red herring.

## Root cause

Two card-build paths pull the ctrl plugin from **two different sources**:

- **Startup / active view** — `R/board-ui.R:36`:
  ```r
  ctrl_ui = if ("ctrl_block" %in% names(plugins)) plugins[["ctrl_block"]]
  ```
  `plugins` here is the **served `serve(plugins=)` set**, which includes `ctrl_block`. → toggle renders.

- **First visit to another view** — `ensure_block_ui` (`R/block-ui.R:204-221`) re-derives from the board:
  ```r
  plugins <- board_plugins(x)
  ...
  ctrl_ui = if ("ctrl_block" %in% names(plugins)) plugins[["ctrl_block"]]
  ```
  `board_plugins.board` (blockr.core `R/board-plugins.R:318`) is a **hardcoded default set that has `edit_block` but NO `ctrl_block`** → `ctrl_ui = NULL` → `block_card` renders edit + eye but **no control toggle**.

`ensure_block_ui`'s own comment says it derives "edit / ctrl UI the way `board_ui` does" — but it can't, because `board_plugins(x)` never carried the served `ctrl_block`. Call chain: `switch_active_view` (`R/dock-view.R:1590`, receives only the board `brd`) → `ensure_block_ui`; the served plugins are simply not in scope there.

The block itself is fine — all blocks statically pass the dock's `has_external_ctrl` (`R/block-ui.R:70`). The **render path drops the plugin**.

## Minimal repro (blockr.core + blockr.dock only)

The sparkle is just core's `ctrl_block` plugin, so a plain `ctrl_block()` reproduces the missing-toggle symptom — no `blockr.ai` needed. Includes a **non-interactive two-line proof** (no Shiny):

[...]

PR #332, opened 2026-07-15
Pad the save snapshot to all board blocks in serialize_board.dock_board

christophsax:
Fixes BristolMyersSquibb/blockr.core#279.

Under deferred construction (core's needed-slot eval gate + `blockr.background_construction_delay = Inf`), off-screen views' block cards are only built on first visit. `serialize_board.dock_board` built the save snapshot from **only the constructed blocks**, but `blockr_ser.blocks` asserts one entry per board block (`length(blocks) == length(x)`) — so "Save as new workflow" on a board with unbuilt blocks aborted with `length(blocks) == length(x) is not TRUE`.

This pads the snapshot to every `board_block_ids(x)`, mirroring core's `serialize_board.board`: an unbuilt block maps to `NULL` and serializes from its constructor scope, so its restored state is **preserved, not blanked**.

**Verified** on a 28-block board (`Adept1_study.json`) with only 5 built: partial save succeeds, and each unbuilt block round-trips byte-identically to a fully static serialization. Also OK with all built / none built (static path).

---
Stacked on `301-drop-bridge-debounce` (#303) — the deferred-construction path that exposes this. Retarget to `main` once BristolMyersSquibb/blockr.dock#303 lands. Sibling lazy-view bug (control toggle missing on off-screen cards) is tracked separately in BristolMyersSquibb/blockr.dock#331.

### DIFF
--- a/R/utils-serdes.R
+++ b/R/utils-serdes.R
@@ -14,6 +14,23 @@ serialize_board.dock_board <- function(x, blocks, id = NULL, dock,
     reval_if
   )
 
+  # Deferred construction (core's needed-slot gate, background_construction_delay)
+  # leaves off-screen blocks unbuilt, so `blocks` only carries the constructed
+  # ones -- a PARTIAL snapshot. blockr_ser.blocks asserts one entry per board
+  # block (`length(blocks) == length(x)`), so a partial snapshot aborts the whole
+  # save with "length(blocks) == length(x) is not TRUE". Pad to every board block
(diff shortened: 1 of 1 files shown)


christophsax:
Closing — not opening a PR yet. The fix lives on branch `301-lazy-view-fixes` (`0850c9f`) and is referenced from BristolMyersSquibb/blockr.core#279; will revisit as a PR later.

Issue #20, opened 2025-10-24
Dock disappears after removing all panels


nbenn:
After all panels have been removed, the dock disappears.

nbenn:
@DivadNojnarg one thing that I find a bit unintuitive: when you start up a dockview instance I assume you have a single empty "group" instance. When you now add a panel and remove it again, you end up in a different place, no? With removal of the last panel you have also removed the group. This is probably what you want if you have n > 1 groups. But maybe for the last group you don't want to completely remove it?

One thing that I intuitively think should be satisfied is: you start empty, add a panel and remove it again and then you are back in exactly the same situation as when you started.

nbenn:
> This is the case.

You are ofc right:

```r
serve(new_dock_board())
```

gives the same as starting with

```r
serve(new_dock_board(extensions = new_edit_board_extension()))
```

and removing the extension panel.

DivadNojnarg:
My suggestions:

- When the last panel is removed, we can show the modal again where all the possible choices are shown (block and extensions). Probably the easiest implementation.
- We prevent the last panel from being removed and we can also show a notification ("at leat one panel has to be there..."). I find this a bit arbitrary but that does not sound too complex to implement.

This is related to your comment:

> Not auto-removing the last group could also work (conceptually), but idk what effort would be required to pull this off (i.e. to somehow modify/override the default dockview behavior).

As we have control over `input[[paste0(dock_id(), "_panel-to-remove")]]`, this is possible to customise the logic for the last panel, without too much effort.

- dockview or blockr.dock offer a placeholder button for an empty dock where clicking shows the modal to add panels. Requires to inject and maintain this button in the dock. Sounds like a bit more hacky.


--- 

As a side note, the modal selectInput does not update when a panel is selected, which means as it is currently not closed after a choice, we are still suggested panels that are in the dock. Alternative would be to close the modal after each choice and updating the select input, the counter being that you can't chain panel creation.

PR #335, opened 2026-07-16
Resolve variadic link inputs to a positional slot

DivadNojnarg:
## Symptom

Multi-input Download/Export via the "Connect ..." sidebar produces files named `1.csv`, `2.csv` (and Excel sheets `1`, `2`) instead of meaningful names. Reported for the ALLM QC dashboard.

## Cause

blockr.core's name-or-position variadic model treats an integer input name as a *named* argument, so `resolve_free_input()`'s generated `"1"`, `"2"`, ... quietly named what should be positional, and blockr.io names download files/sheets by the link input verbatim.

blockr.ui fixed this on Jul 1 (`2945271`, #20), but the Jul 13 CRAN vendoring (`4bf6b57`) copied a pre-fix ui tree, silently reverting it in the dock copy. That fix was the only post-Jul-1 change to the five vendored files, so nothing else was lost.

## Change

- `resolve_free_input()` emits an empty (positional) slot for variadic targets, byte-for-byte the ui fix. Covers the link-menu commit and both sidebar-block auto-link call sites.
- Unit tests: sidebar commit resolves a variadic target to `""`; a target already carrying a positional link never gets a generated integer name.

## Verification

Live board (2 datasets -> Download, positional links): CSV ZIP entries `data_1.csv` + `data_2.csv`, Excel sheets `data_1`, `data_2` (fallback names from the companion blockr.io PR cynkra/blockr.io: fill positional slot names). Full suite: 7280 passed, 0 failed.

## Note for review

This restores *correct* (positional) slots but not *meaningful* names: those need either a settable name field in the link menu (user-named slots are legit in core's model) or new core surface exposing slot origins — positional slots reach the block server with hidden synthetic keys, so the upstream block id is not available to blockr.io. Decision pending with @christoph.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # blockr.dock (development version)
 
+* `resolve_free_input()` now resolves a variadic link target to an empty
+  (positional) slot instead of a generated `"1"`, `"2"`, ... name, aligning
+  the "Connect ..." sidebar with blockr.core's name-or-position variadic
+  model (and with blockr.ui's link menu). An integer input name is treated
+  by core as a *named* argument, and downstream consumers such as
+  blockr.io's Download/Export blocks name files and Excel sheets by the
+  link input verbatim, which surfaced as `1.csv` / `2.csv` exports.
+
 * At startup the board now builds only the active view's dockView;
   off-screen views' docks are created on first visit rather than all up
   front (#304), mirroring the card deferral below. Building every view's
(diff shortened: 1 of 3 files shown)


nbenn (approved):
LGTM — correct, minimal fix. `resolve_free_input()` now resolves a variadic target to an empty positional slot instead of a synthetic `1`/`2` name, matching blockr.core's name-or-position model. I confirmed that two positional links into one variadic block validate against core (empty inputs are exempt from the dup-target check via `filter_empty()`), that the new tests fail on the pre-fix code, and that the finite-arity path is untouched. Non-blocking test-quality follow-ups (a test comment that labels a named `"1"` input as "positional"; missing coverage for the already-carries-a-positional-link steady state) are folded into the meaningful-names tracking issue #337 rather than blocking here.

PR #365, opened 2026-07-22
Keep the navbar busy spinner turning under reduced motion

christophsax:
The reduced-motion block dropped the ring's animation outright. Since the spinner is hidden when idle, "frozen" was the only state those users ever saw it in: a fully styled ring sitting still in the navbar, reading as a hung session rather than a busy one.

Windows maps its "Animation effects: off" setting -- a common managed / VDI default -- onto prefers-reduced-motion: reduce, so this hit every Edge user on such a machine while spinning fine on a developer laptop.

Slow the turn to 1.6s instead of stopping it. A 16px ring turning in place is not the page-level motion the preference exists to suppress.

Guarded by a stylesheet assertion, because nothing else can catch it: the spinner is only ever seen mid-flush and CI never runs with the preference set.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,13 @@
   as the free-text summary; the earlier `extension_description()` accessor is
   deprecated in favour of `ext_desc()` (#359).
 
+* The navbar busy spinner keeps turning when the browser reports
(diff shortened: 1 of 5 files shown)


christophsax:
Closes #371

christophsax:
Closes #372

PR #378, opened 2026-07-24
Adopt core's logical visible channel as the card-build ledger

### LINKED ISSUE #377: "Show code" permanently blanks unvisited views on a deferred dock board
## Agreed approach (supersedes the "Possible fixes" below and my earlier option-2 comment)

This is fixed by **repurposing an existing core channel, not by adding a dock-side ledger**. The diagnosis below is accurate; only the fix direction changed.

Root cause, one level deeper: core's `visible` channel is a two-state string (`NA_character_` / view-id) read as `is_visible <- !is.na`, so it cannot distinguish "never built" from "built but off-screen" — the one fact a card-build ledger needs. Rather than bolt a parallel dock axis onto the side, we make `visible` a logical three-state channel, symmetric with `required`. That is blockr.core#306 (PR blockr.core#307):

- `visible`: `NA` never-built / `FALSE` built-off-screen / `TRUE` painted-now
- `is_visible <- isTRUE(visible)` — painted now (the render gate)
- `!is.na(visible)` — ever built — the dock's build ledger

The dock change here is an **adoption, with no new channel**:

- drop the `visibility$built` axis entirely
- `built_cards() <- !is.na(visibility$visible[[id]]())`
- `mark_cards_rendered` writes `TRUE` (not the view-id string); `show_cards` writes `FALSE` off-screen (not `NA_character_`) — that `FALSE` is the persistence that fixes this bug; the startup seed writes `FALSE` for the active view's built cards

Lands together with blockr.core#307 via a revdep pin (the blockr.core#273 / #274 rollout pattern). **PR #378 in its current form — the `visibility$built` axis plus view-id string writes — is the pre-pivot approach and will be reworked to the above, not merged standalone.**

---

# "Show code" permanently blanks unvisited views on a deferred dock board

`require_all_blocks()` (plugin-code.R) marks every block
`visibility$required[[id]](TRUE)` and never withdraws the demand. That is
correct for construction, but blockr.dock uses the *same* `required` axis as
its card build ledger: `built_cards()` treats any non-NA slot as "card
built". So on a board with `background_construction_delay = Inf`, opening
"Show code" before visiting a view flips that view's (card-less) blocks to
`TRUE`, and the first visit to the view then skips the card build
(`build_block_ui()` computes `new = setdiff(members, built_cards(...))`).
The panels exist, the bodies stay empty, and nothing ever builds them.

This is a follow-on to BristolMyersSquibb/blockr.core#300 / BristolMyersSquibb/blockr.core#303: the export side is fixed (the modal
shows correct code now, "Preparing code..." while blocks construct), but
the demand it plants outlives its purpose and corrupts the container's
ledger.

## Reproduction (core + dock mains, default plugins only)

```r
options(blockr.background_construction_delay = Inf)

library(blockr.core)
library(blockr.dock)

serve(new_dock_board(
  blocks = c(
    a = new_dataset_block("iris"),
    b = new_dataset_block("mtcars"),
    c = new_head_block()
  ),
  links = links(from = "b", to = "c"),
  views = list(Main = "a", Extra = c("b", "c"))
))
```

* Control: start the app, switch to **Extra** -> both cards render
  (mtcars table, head). :white_check_mark:
* Bug: start a fresh session, click **Show code** first (modal shows the
  full script -- the BristolMyersSquibb/blockr.core#303 fix working), close it, switch to **Extra** ->
  the "Dataset" / "Head" panel tabs are there but both bodies are blank,
  permanently. :x:

Verified against blockr.core fd26bcf + blockr.dock f008e02 (headless
chromium; screenshots `showcode-control.png` / `showcode-bug.png`).

## Possible fixes

1. **Withdraw the demand** (smallest): `require_all_blocks()` snapshots
   each slot's prior value and restores it once the export has been
   served (the code render completing is the natural "done" point). A
   constructed server behind an unbuilt card is a state the dock already
   supports -- finite-delay background construction produces it routinely
   -- so restoring NA/FALSE leaves core with the built block and the dock
   with an honest ledger. (blockr.outline's report download uses exactly
   this pattern against the forwarded visibility channel.)
2. **Stop overloading `required`**: give the dock its own build-ledger
   axis so a demand can never masquerade as a built card. More invasive,
   but removes the trap for every future `required` writer.
3. **First-class demand API**: a core-level `demand_blocks(ids)` /
   `release_blocks(ids)` pair that containers implement, so consumers
   (generate_code, report surfaces) never touch the raw slots.

Option 1 fixes the bug today; option 2 or 3 is the structural cure.

## Related

- Follow-on to BristolMyersSquibb/blockr.core#300 / BristolMyersSquibb/blockr.core#303 (export side already fixed).
- The forwarding seam for the option-1 fix already exists on
  [`BristolMyersSquibb/blockr.dock@feat/visibility`](https://github.com/BristolMyersSquibb/blockr.dock/tree/feat/visibility)
  (commit `6e9b793`, "Forward the visibility channel to extension servers"):
  it passes core's `visibility` bundle through to dock extension servers, so
  a consumer can demand blocks *with* snapshot/restore of the slots — the
  pattern blockr.outline's report download uses, and what avoids exactly this
  corruption.

nbenn:
## Summary

- On a deferred board (`background_construction_delay = Inf`), opening "Show code" before visiting a view left that view's cards permanently blank on the first visit.
- Root cause: `built_cards()` read the block's `required` slot as the dock's build ledger, but `required` is a multi-writer construction-demand channel. Core's "Show code" (`require_all_blocks`) marks every block required to export the whole script, so an off-screen block's demand read back as a built card and the view switch skipped its build.
- Reworked to the agreed cross-package fix, superseding the earlier `visibility$built` dock-side ledger: blockr.core#306 makes `visible` a logical three-state channel — `NA` never built / `FALSE` built off screen / `TRUE` painted, with `is_visible <- isTRUE` — and the dock folds its build ledger onto `!is.na(visible)` with no parallel axis.
- Dock changes: `built_cards()` reads `!is.na(visible)`; `mark_cards_rendered` writes `TRUE`; `show_cards` writes `FALSE` off screen — the persistence that fixes the bug, since a card leaving the screen keeps its ledger entry instead of being erased to `NA`; and the startup seed marks the active view's cards built (`FALSE`). Core's `require_all_blocks` is untouched.
- blockr.core#306 (the logical `visible` channel) has now merged to `main`, so the interim `blockr.core@306-visible-logical` pin has been dropped — this depends on plain `blockr.core`. Rebased onto current `main`.

Fixes #377.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,17 @@
 # blockr.dock (development version)
 
+* "Show code" on a deferred board
+  (`background_construction_delay = Inf`) no longer permanently blanks
+  the cards of a view first visited after it. The exporter marks every
+  block required so the script covers the whole board, and the dock's
+  card ledger -- read off that same `required` channel -- mistook the
+  export's demand for a built card, so the first visit to the view
+  skipped the build and left its panels empty. The dock now reads its
+  build ledger off core's `visible` channel, which blockr.core makes a
+  logical three-state axis (`NA` never built / `FALSE` built off screen
+  / `TRUE` painted), so a `required` write can no longer masquerade as a
+  built card (#377).
+
 * New `edit_inputs_action`: a per-block sidebar listing every incoming link
   as an ordered list, so a block's inputs can be managed together rather than
   one edge at a time. For a variadic block each row is a positional slot -
--- a/R/block-ui.R
+++ b/R/block-ui.R
@@ -121,38 +121,42 @@ insert_block_ui.dock_board <- function(id, x, blocks, dock, ...,
   invisible(x)
 }
 
-# The dock's build ledger and per-block visibility both live on the `visibility`
-# channel core hands the board callback: two per-block reactiveVal environments
-# (`required`, `visible`), one slot per board block, all core-owned. A card is
-# built once the dock writes its `required` slot (non-NA), so `built_cards()`
-# reads the built set straight off it -- there is no second ledger. The dock
-# writes `required[[id]]` FALSE when it builds a card (off screen) and TRUE when
-# on screen, and `visible[[id]]` the view id once the client paints it. Block
-# removal needs no dock write: core drops the slot, dropping the card too.
-
-# The blocks whose cards are built: those with a non-NA required slot.
+# The dock's build ledger is core's `visible` axis: a per-block reactiveVal,
+# logical (NA never built / FALSE built off screen / TRUE painted now) like
+# `required`, so `built_cards()` reads `!is.na(visible)`. It once read
+# `required` non-NA, but that axis is a multi-writer construction-demand
+# channel -- core's "Show code" marks every block required to export the whole
+# script -- so a demand with no card masqueraded as built and blanked the view
+# on its first visit. `visible` is written only where the dock builds, paints
+# or reconciles a card. The dock still writes `required[[id]]` FALSE off screen
+# / TRUE on screen (construction demand); block removal needs no dock write --
+# core drops the slot, dropping the card from the ledger too.
 built_cards <- function(visibility) {
-  ids <- ls(visibility$required)
-  ids[lgl_ply(ids, slot_built, visibility$required)]
+  ids <- ls(visibility$visible)
+  ids[lgl_ply(ids, slot_built, visibility$visible)]
 }
 
-slot_built <- function(id, required) {
-  !is.na(isolate(required[[id]]()))
+slot_built <- function(id, visible) {
+  !is.na(isolate(visible[[id]]()))
 }
 
-# Record `new` cards as built: their required slot goes FALSE (off screen until
-# the report observer places them). Slots pre-exist -- core seeds every block's
-# slots before any block plugin runs.
+# Record `new` cards as built off screen: `visible` FALSE enters them in the
+# ledger (built, not yet painted), `required` FALSE holds no construction
+# demand until a view switch or the report observer places them. Slots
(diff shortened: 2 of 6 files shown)


nbenn:
## Hold — superseded by blockr.core#306 / blockr.core#307, rework rather than merge

This PR takes the pre-pivot approach: it adds a `visibility$built` parallel axis and still writes view-id strings into `visible`. We've since decided to fix the root instead — make core's `visible` a logical three-state channel (`NA` never / `FALSE` built-off-screen / `TRUE` painted-now; `is_visible <- isTRUE`), so the dock reads `!is.na(visible)` as the build ledger with no parallel axis. Reasoning and source-cited diff in blockr.core#306.

Rework to the adoption (do not merge as-is):

- drop `visibility$built` (and its seed / prune / `mark_cards_built` write)
- `built_cards() <- !is.na(visibility$visible[[id]]())`
- `mark_cards_rendered` → writes `TRUE` (not the view-id string)
- `show_cards` off-screen → writes `FALSE` (not `NA_character_`) — this `FALSE` is the persistence that fixes the bug
- startup seed → `FALSE` for the active view's built cards
- lands with blockr.core#307 via a revdep pin (blockr.core#273 / #274 pattern)

Keep the regression tests as-is — they assert the contract (a bare `required` write, i.e. "Show code", must not suppress the build; and the first visit to a deferred view builds its cards). Only the mechanism underneath changes from the `built` axis to `!is.na(visible)`.

